PR #7, opened 2018-05-14
Escape backslash in README

jobegrabber:


dwelch-spike:
Looks like this change was done in https://github.com/aerospike/aerospike-tools-backup/commit/f6ace8fc65f0e8869f70b0456e39c9a83995f3ca but this PR was never closed. Sorry this was never properly addressed. Thanks for the contribution. Closing now.

PR #10, opened 2022-04-20
[TOOLS-1996] (ASBACKUP) Build asbackup for macOS in github actions.

ClaytonKnittel:


jdogmcsteezy (changes requested):
I think it needs to be split into a separate file.  In the future, it will likely need to be separate anyway

.github/workflows/backup-artifact.yml
```diff
@@ -1,17 +1,31 @@
...
+        if [ "$RUNNER_OS" == "Linux" ]; then
+          sudo apt install -y libcurl4-openssl-dev libssl-dev uuid-dev zlib1g-dev libpulse-dev
+        elif [ "$RUNNER_OS" == "macOS" ]; then
+          brew update
+          brew install openssl@1.1 readline
+          brew install cmake
+        else
+          echo "$RUNNER_OS not supported"
+          exit 1
+        fi
```
jdogmcsteezy: The aws cache below does not look correct.  If you changed the `ref:` to the aws SDK wouldn't it still have a cache hit?  The example [here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) hashes the files to make sure they are indeed the same.

ClaytonKnittel: I have to disagree with this. I would expect anyone modifying this file to know that the cache key has to change if they want to change the ref. We can add the tag to the cache key if that would make it clearer.

We also skip the downloading of the repo if there is a cache hit, so if we wanted to use a hash of some set of files from the repo to determine the key, we'd have to download it first. aws-sdk-cpp is very very large, so downloading it takes a long time, but restoring the cache is fast.

ClaytonKnittel:
> I think it needs to be split into a separate file. In the future, it will likely need to be separate anyway

Do you mean we should split the mac artifact and linux artifact actions into multiple actions?

PR #11, opened 2022-04-28
[TOOLS-1996] (ASBACKUP) Build asbackup for macOS in github actions.

ClaytonKnittel:


jdogmcsteezy (approved):
I know we talked about somehow caching the artifact before the aws cache is restored.  I was thinking you could somehow cache on aws version and commit of repo.  Maybe even add an extra shallow checkout of the asbackup repo.  It would save a bunch of time when debugging/improving the build process since we won't have to wait so long for asbackup.  Again, I think we talked about it and there were valid reasons not to this

PR #22, opened 2022-08-17
Tools 2092 (asbackup) uninitialized record check before move records

atefehmohseni:


src/utils.c
```diff
@@ -1798,7 +1798,13 @@ bool
 as_key_move(as_key* dst, as_key* src)
 {
 	*dst = *src;
-	if (as_load_uint32(&src->value.integer._.count) > 1) {
+	if (!src->valuep) {
+		return true;
+	}
+	if (as_load_uint32(&src->valuep->integer._.count) > 1 || \
+		as_load_uint32(&src->valuep->string._.count) > 1 || \
+		as_load_uint32(&src->valuep->bytes._.count) > 1 ) {
```
dwelch-spike: Per Brian's information lets change this back to a single check on integer._.count.

test/unit/test_utils.c
```diff
@@ -367,6 +367,73 @@ START_TEST(test_timespec_diff_1ms_overflow)
 }
 END_TEST
 
+
+START_TEST(test_as_key_move_an_empty_rec)
```
dwelch-spike: nit: "empty_rec" should maybe be "empty_key"

test/unit/test_utils.c
```diff
@@ -367,6 +367,73 @@ START_TEST(test_timespec_diff_1ms_overflow)
...
+{
+	as_key* test_key = (as_key *) cf_malloc(sizeof(as_key));
+	
+	ck_assert(as_key_move(test_key, test_key));
+}
+END_TEST
+
+START_TEST(test_as_key_move_a_not_empty_int_key)
```
dwelch-spike: nit: maybe "not_empty" could be "populated"

test/unit/test_utils.c
```diff
@@ -367,6 +367,73 @@ START_TEST(test_timespec_diff_1ms_overflow)
 }
 END_TEST
 
+
+START_TEST(test_as_key_move_an_empty_rec)
+{
+	as_key* test_key = (as_key *) cf_malloc(sizeof(as_key));
```
dwelch-spike: I see Clayton had defined a `test_malloc` function that is used in other tests. I think we should use that here. I think the memory allocated by these should also be freed in the same tests. We have a common citrusleaf file called alloc.h that defines wrapped versions of these prefixed with `cf_`. I think using `cf_free` would be good. So `test_malloc()` and `cf_free()`.

src/utils.c
```diff
@@ -1798,7 +1798,11 @@ bool
 as_key_move(as_key* dst, as_key* src)
 {
 	*dst = *src;
-	if (as_load_uint32(&src->value.integer._.count) > 1) {
+	if (!src->valuep) {
+		return true;
+	}
+	if (as_load_uint32(&src->valuep->integer._.count) > 1) {
```
dwelch-spike: nit: we usually put an empty line between if statements. Unless there is something else specified in Clayton's style guide.

dwelch-spike (changes requested):
I have a few comments and suggestions. Let me know what you think.

test/unit/test_utils.c
atefehmohseni: thanks for the info. Could you please check out the order of calling "cf_free" in the latest commit?

dwelch-spike (approved):
LGTM but lets wait for actions tests to finish before merging.

PR #26, opened 2022-08-26
TOOLS-2121 (ASBACKUP) warn and skip backing up secondary indexes with…

atefehmohseni:
… context

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -1509,6 +1509,7 @@ parse_index_info(char *ns, char *index_str, index_param *index)
 	index->set = NULL;
 	index->name = NULL;
 	index->type = INDEX_TYPE_INVALID;
+	index->ctx = false;
 	as_vector_init(&index->path_vec, sizeof (path_param), 25);
 
 	char *path = NULL;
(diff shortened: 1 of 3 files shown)


src/utils.c
```diff
@@ -1564,8 +1565,10 @@ parse_index_info(char *ns, char *index_str, index_param *index)
 				err("Invalid index type %s", arg);
 				goto cleanup2;
 			}
-		} else if (strcmp(para, "path") == 0 || strcmp(para, "context") == 0) {
+		} else if (strcmp(para, "bin") == 0) {
 			path = arg;
+		} else if (strcmp(para, "context") == 0 && strcasecmp(arg, "NULL") != 0) {
```
dwelch-spike, line 1570: question for my understanding. What necessitates the ` && strcasecmp(arg, "NULL") != 0)`?

atefehmohseni: all the records for `sindex-list` command have a `context` parameter but only those with a not null value should be skipped.

dwelch-spike (approved):
LGTM

PR #28, opened 2022-08-31
Tools 2088 - ASBACKUP/ASRESTORE secondary index(es) with ctx

atefehmohseni:
for integration test, build the python client from 'dev' branch.

currently, c-client module is on 'stage' branch (to call the conversion b64 encoded ctx <> cdt_ctx func.)


dwelch-spike:
@atefehmohseni please change the target branch to test-ready. If it does not exist you can create it from master.

I see the tests failed with a segfault for the last commit. What's going on there?

atefehmohseni:
The problem with the unit test was resolved. I needed to close the backup file before opened it again in `encode/decode` backup file.

PR #29, opened 2022-09-01
[TOOLS-2137] Fix uploader leak.

dwelch-spike:


atefehmohseni (commented):
Looks like `_batch_tracker_destroy` function also only destroy the reference pointer to `tracker->records` and not the individual records. It does free records inside the `tracker->batch`  though (aka `_free_batch_records` )
Could you please clarify why the same for loop to free each record inside a  tracker is not needed at `_batch_tracker_destroy`?

dwelch-spike:
> 

At first I thought it was not needed because the ops that are freed by _free_batch_records free the same bins that the records use. This isn't all the memory though. Throwing the record destroy after the _free_batch_records ensures the bin values have the correct reference count and that all memory is freed.

atefehmohseni (approved):
LGTM

PR #32, opened 2022-09-08
[Tools 2088] backup/restore Sindex with ctx

atefehmohseni:


atefehmohseni:
Comments on the same PR:

https://github.com/aerospike/aerospike-tools-backup/pull/28#issuecomment-1240017329

https://github.com/aerospike/aerospike-tools-backup/pull/28#issuecomment-1241053849

src/dec_text.c
```diff
@@ -1548,13 +1548,23 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
...
 
-		if (!expect_char(fd, line_no, col_no, i == n_paths - 1 ? '\n' : ' ')) {
+		if (!expect_char(fd, line_no, col_no, ' ')) {
 			goto cleanup3;
 		}
 
 		as_vector_append(&index->path_vec, &path);
 	}
 
+	if (!text_nul_read_token(fd, false, line_no, col_no, ctx, sizeof ctx, "\n")) {
```
atefehmohseni, line 1558: add tests to call old backup files (that dont have ctx at the end of sindex records)

src/enc_text.c
```diff
@@ -515,8 +515,8 @@ text_put_secondary_index(io_write_proxy_t *fd,
 			return false;
 		}
 	}
-
-	if (io_proxy_printf(fd, "\n") < 0) {
+	
+	if (io_proxy_printf(fd, " %s\n", index->ctx != NULL ? escape(index->ctx) : "") < 0) {
```
atefehmohseni: separate the "\n" write to be explicit

src/backup.c
```diff
@@ -1509,11 +1509,11 @@ process_secondary_indexes(backup_job_context_t *bjc)
 	ver("Processing secondary indexes");
 
 	bool res = false;
-
-	size_t value_size = sizeof "sindex-list:ns=" - 1 + strlen(bjc->conf->ns) + 1;
+	char* b64_enable = ";b64=true";
```
atefehmohseni, line 1512: test if older server versions would take "b64=true"

src/utils.c
```diff
@@ -1565,12 +1566,11 @@ parse_index_info(char *ns, char *index_str, index_param *index)
 				err("Invalid index type %s", arg);
 				goto cleanup2;
 			}
-		} else if (strcmp(para, "bin") == 0) {
```
atefehmohseni, line 1568: consider "path" apart from "bin" from sindex-list server response.

src/restore.c
```diff
@@ -1061,9 +1061,10 @@ check_index(aerospike *as, index_param *index, uint32_t timeout)
 
 	index_status res = INDEX_STATUS_INVALID;
 
-	size_t value_size = sizeof "sindex-list:ns=" - 1 + strlen(index->ns) + 1;
```
atefehmohseni, line 1064: "b64=true" with older server test -> asrestore behavior?

src/restore.c
```diff
@@ -1168,6 +1169,13 @@ check_index(aerospike *as, index_param *index, uint32_t timeout)
 		}
 	}
 
+	if (!index->ctx && index->ctx != index2.ctx) {
```
atefehmohseni: checking `!index->ctx` is extra
user `strcmp` for comparison

src/restore.c
```diff
@@ -1314,15 +1322,31 @@ restore_index(aerospike *as, index_param *index, as_vector *set_vec,
...
 				index->set[0] == 0 ? NULL : index->set, path->path, index->name, itype,
-				dtype) != AEROSPIKE_OK) {
+				dtype, index->ctx == NULL ? NULL: &ctx) != AEROSPIKE_OK) {
 		err("Error while creating index %s:%s:%s (%s) - code %d: %s at %s:%d", index->ns,
 				index->set, index->name, path->path, ae.code, ae.message, ae.file, ae.line);
+		
+		as_cdt_ctx_destroy(&ctx);
 		return false;
 	}
-
```
atefehmohseni, line 1325: destroy cdt_ctx if no error occurs too

test/unit/test_utils.c
```diff
@@ -437,13 +437,35 @@ START_TEST(test_as_key_move_success)
...
 	ck_assert(as_key_move(dst, src));
-	ck_assert(dst->value.integer.value == src->value.integer.value);
+	ck_assert_int_eq(dst->value.integer.value, src->value.integer.value);
 	cf_free(test_key_value);
 	cf_free(dst);
 	cf_free(src);
 }
 END_TEST
 
+START_TEST(test_secondary_index_param_with_ctx)
```
atefehmohseni, line 447: add a unit test to sindex record without "context" in server response (index->ctx be null)

atefehmohseni (commented):
review comments from the review session with Jesse and Dylan

src/restore.c
```diff
@@ -52,8 +52,8 @@ static bool open_file(const char *file_path, as_vector *ns_vec, io_read_proxy_t
 static bool check_set(char *set, as_vector *set_vec);
 static void * restore_thread_func(void *cont);
 static void * counter_thread_func(void *cont);
-static const char * print_set(const char *set);
-static bool compare_sets(const char *set1, const char *set2);
+static const char * print_optional_str(const char *set);
+static bool compare_strs(const char *set1, const char *set2);
```
dwelch-spike: the parameters are still names `set` and `set2` despite being changed in the definition.

dwelch-spike: Params still called `set1` and `set2`

src/restore.c
```diff
@@ -1168,6 +1169,13 @@ check_index(aerospike *as, index_param *index, uint32_t timeout)
 		}
 	}
 
+	if (!compare_strs(index->ctx, index2.ctx)) {
```
dwelch-spike, line 1172: With the new check, won't this enter the body of the if statement if both `index->ctx` and `index2.ctx` are NULL (null pointers)? Is that expected?

src/utils.c
```diff
@@ -1806,7 +1806,7 @@ as_key_move(as_key* dst, as_key* src)
 	}
 	
 	if (as_load_uint32(&src->valuep->integer._.count) > 1) {
-		inf("Couldn't move record key values (reference count > 1).");
+		//inf("Couldn't move record key values (reference count > 1).");
```
dwelch-spike, line 1809: Should this be commented?

dwelch-spike (changes requested):
A few comments/questions. Looks good otherwise.

src/restore.c
dwelch-spike: duplicate

PR #35, opened 2022-10-07
[TOOLS-2172] Linux arm64 support.

dwelch-spike:


src/io_proxy.c
```diff
@@ -916,18 +916,10 @@ _consumer_buffer_free(consumer_buffer_t* cb)
...
-	__asm__("cmp %[pos], %[dpos]\n\t"
-			"cmove %[zero], %[pos]\n\t"
-			"cmove %[zero], %[dpos]\n\t"
-			: [pos] "+r" (pos),
-			  [dpos] "+r" (dpos)
-			: [zero] "r" (zero)
-			: "cc");
-	*pos_ptr = pos;
-	*dpos_ptr = dpos;
+	if (*pos_ptr == *dpos_ptr) {
```
jdogmcsteezy, line 919: Well, that makes sense now.  I can't imagine there is a change in performance here.  I am not sure what line 925->927 were doing, though.

jdogmcsteezy (approved):
LGTM, we talked about some extra testing/tinkering you could do to verify and explain what the heck that assembly code is saying.  We know what it is doing but are not sure why the old code used the methods it did

src/io_proxy.c
dwelch-spike: I think the point was to do a conditional move without the branch prediction penalty normally generated by an if.

dwelch-spike:
> LGTM, we talked about some extra testing/tinkering you could do to verify and explain what the heck that assembly code is saying. We know what it is doing but are not sure why the old code used the methods it did

I ran some backups and restores using asbackup master vs this branch. Everything seems fine, even when encrypting with the old tool and decrypting with the new one. 

I also tried flipping htobe64 and be64toh like we discussed, the unit tests still pass. It doesn't seem to make a difference.

PR #37, opened 2022-10-22
[TOOLS-2228] gcc plugin

dwelch-spike:
Use the gcc arm barrier plugin.
TODO test and change for mac builds.


kportertx:
You shouldn't commit the binary `.so` to git.

dwelch-spike:
> 

Removed and updated gitignore.

kportertx:
Be sure to squash and merge so that the inclusion of the binary is scrubbed from the history and doesn't make certain git operations (such as cloning) slower in the future.

src/batch_uploader.c
```diff
@@ -1058,7 +1058,8 @@ _key_put_submit_callback(as_error* ae, void* udata, as_event_loop* event_loop)
 			break;
 	}
 
-	if (as_aaf_uint64(&tracker->outstanding_calls, -1lu) == 0) {
+	if (as_aaf_uint64_rls(&tracker->outstanding_calls, -1lu) == 0) {
+		as_fence_seq();
```
jdogmcsteezy: Could be made a `acq`.  We might want to add comments about why these are needed as well and in what case we can remove these memory barriers.  Like, if we switch the c11 I assume this will add redundant barriers.

src/batch_uploader.c
```diff
@@ -1103,10 +1104,11 @@ _do_key_recs_write(batch_uploader_t* uploader, record_batch_tracker_t* tracker)
 				// number that failed to initialize (this one and all succeeding
 				// ones). If we happen to decrease this value to 0, free the
 				// tracker and release our hold on an async batch slot.
-				if (as_aaf_uint64(&tracker->outstanding_calls,
+				if (as_aaf_uint64_rls(&tracker->outstanding_calls,
 							(uint64_t) -(n_records - i)) == 0) {
 					// if this is the last record, we can make the upload_batch
 					// callback.
+					as_fence_seq();
```
jdogmcsteezy: same

jdogmcsteezy (changes requested):

src/batch_uploader.c
dwelch-spike: Changed to acquire fences. Left a comment above one of them.

src/batch_uploader.c
dwelch-spike: changed

src/batch_uploader.c
jdogmcsteezy: I think something more explicit like an inline `// remove after switching to c11` or something but I think it is fine

jdogmcsteezy (approved):
LGTM

PR #39, opened 2022-11-19
asbackup 3.13.1 [TOOLS-2263] unable to restore backup file from pre 3.12.0

dwelch-spike:
Fix interpretation of older backup file sindexs that don't append a space.

### DIFF
--- a/src/dec_text.c
+++ b/src/dec_text.c
@@ -1508,6 +1508,9 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
 	as_vector_init(&index->path_vec, sizeof (path_param), 25);
 	path_param path;
 
+	// This loop parses n_path path specs from sindex entries in the
+	// backup file. These include the space after n_paths to the path type.
+	// The code supports multiple path specs, but currently there is only ever one.
 	for (size_t i = 0; i < n_paths; ++i) {
 		if (!expect_char(fd, line_no, col_no, ' ')) {
 			goto cleanup2;
@@ -1548,6 +1551,8 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
 			goto cleanup3;
 		}
 
+		// If this isn't the last path spec, the only valid seperator is ' '
+		// NOTE: currently there is only ever 1 path spec.
 		if (i != n_paths - 1 && !expect_char(fd, line_no, col_no, ' ')) {
 			goto cleanup3;
 		}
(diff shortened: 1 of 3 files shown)


src/dec_text.c
```diff
@@ -1508,6 +1508,9 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
 	as_vector_init(&index->path_vec, sizeof (path_param), 25);
 	path_param path;
 
+	// This loop parses n_path path specs from sindex entries in the
+	// backup file. These include the space after n_paths to the path type.
+	// The code supports multiple path specs, but currently there is only ever one.
```
jdogmcsteezy:
```suggestion
	// This loop parses n_path path specs from sindex entries in the
	// backup file. These include the space after n_paths to the path type.
	// The code supports multiple path specs, but there is only ever one.
	// paths should be changed to a single value rather than a vector to
	// avoid future confusion.
```

src/dec_text.c
```diff
@@ -1548,13 +1551,26 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
...
+		// NOTE: currently there is only ever 1 path spec.
+		if (i != n_paths - 1 && !expect_char(fd, line_no, col_no, ' ')) {
 			goto cleanup3;
 		}
 
 		as_vector_append(&index->path_vec, &path);
 	}
 
+	// If this is a pre asbackup 3.12.0 (has not whitespace after final path spec)
+	// then leave the newline to delimit the ctx.
```
jdogmcsteezy:
```suggestion
	// then leave the newline to terminate the ctx. 
```

src/dec_text.c
```diff
@@ -1548,13 +1551,26 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
...
+		// If this isn't the last path spec, the only valid seperator is ' '
+		// NOTE: currently there is only ever 1 path spec.
+		if (i != n_paths - 1 && !expect_char(fd, line_no, col_no, ' ')) {
 			goto cleanup3;
 		}
 
 		as_vector_append(&index->path_vec, &path);
 	}
 
+	// If this is a pre asbackup 3.12.0 (has not whitespace after final path spec)
```
jdogmcsteezy:
```suggestion
	// If this is a pre asbackup 3.12.0 (has no whitespace after final path spec)
```

src/dec_text.c
```diff
@@ -1548,13 +1551,23 @@ text_parse_index(io_read_proxy_t *fd, as_vector *ns_vec, uint32_t *line_no,
...
+		// If this isn't the last path spec, the only valid seperator is ' '
+		// NOTE: currently there is only ever 1 path spec.
+		if (i != n_paths - 1 && !expect_char(fd, line_no, col_no, ' ')) {
 			goto cleanup3;
 		}
 
 		as_vector_append(&index->path_vec, &path);
 	}
 
+	// If this is asbackup 3.12.0 (has whitespace after final path spec)
```
jdogmcsteezy:
```suggestion
	// If this is asbackup 3.12.0 or greater and sindex has a ctx then consume the delim
```

src/enc_text.c
```diff
@@ -515,11 +515,14 @@ text_put_secondary_index(io_write_proxy_t *fd,
...
-		err("Error while writing secondary index to backup file [3]");
-		return false;
+
+	if (index->ctx != NULL) {
+		if (io_proxy_printf(fd, " %s", escape(index->ctx)) < 0) {
+			err("Error while writing secondary index to backup file [3]");
+			return false;
+		}
 	}
+
```
jdogmcsteezy:
```suggestion
```

PR #43, opened 2022-12-09
simplify mac build

dwelch-spike:
Removed aws and asbackup dynamic builds. Install cmake via brew.

### DIFF
--- a/.github/workflows/backup-artifact-linux.yml
+++ b/.github/workflows/backup-artifact-linux.yml
@@ -1,7 +1,7 @@
 name: Build:Main
 on:
   push:
-    branches: [ master, test-ready ]
+    branches: [ master ]
 jobs:
   build-libraries:
     runs-on: ubuntu-20.04
--- a/.github/workflows/mac-artifact.yml
+++ b/.github/workflows/mac-artifact.yml
@@ -2,6 +2,8 @@ name: Mac Artifact
 on:
   push:
     branches: [ master, test-ready ]
+  pull_request:
+    branches: [ master, test-ready ]
   workflow_call:
     inputs:
       submodule:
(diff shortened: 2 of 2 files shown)


jdogmcsteezy (changes requested):
LGTM, but I would change the mac-workflow to build on PR into test-ready in order to test

.github/workflows/backup-artifact-linux.yml
```diff
@@ -1,7 +1,7 @@
 name: Build:Main
 on:
   push:
-    branches: [ master, test-ready ]
+    branches: [ master ]
```
jdogmcsteezy, line 4: you sure you don't want the extra check?

.github/workflows/mac-artifact.yml
```diff
@@ -2,6 +2,8 @@ name: Mac Artifact
 on:
   push:
     branches: [ master, test-ready ]
```
jdogmcsteezy:
```suggestion
    branches: [ master, test-ready, "bugfix-*", actionsHub ]
```
I find both these useful.  The bugfix one is necessary I think

.github/workflows/mac-artifact.yml
```diff
@@ -42,33 +43,9 @@ jobs:
...
-      if: steps.cache-cmake.outputs.cache-hit != 'true'
-      run: |
-        ./bootstrap 
-        make
-      working-directory: CMake-${{ env.CMAKE_VERSION }}
-    - name: install cmake
-      run: |
-       sudo make install
-      working-directory: CMake-${{ env.CMAKE_VERSION }}
+        brew install openssl@1.1 readline cmake
```
jdogmcsteezy, line 46: Do we still need 1.1?

jdogmcsteezy (approved):
some questions but will approve

.github/workflows/backup-artifact-linux.yml
dwelch-spike: This workflow sets the latest tag. I don't want it set to any commits on test-ready.

PR #47, opened 2023-01-09
TOOLS-2309 fix: support restoring more than 1k objects per s3 bucket

dwelch-spike:
Also update copyright and python client test dependency.

### DIFF
--- a/src/file_proxy_s3.cc
+++ b/src/file_proxy_s3.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Aerospike, Inc.
+ * Copyright 2023 Aerospike, Inc.
  *
  * Portions may be licensed to Aerospike, Inc. under one or more contributor
  * license agreements.
@@ -31,7 +31,7 @@
 #include <aws/s3/model/AbortMultipartUploadRequest.h>
 #include <aws/s3/model/HeadObjectRequest.h>
 #include <aws/s3/model/ListMultipartUploadsRequest.h>
-#include <aws/s3/model/ListObjectsRequest.h>
+#include <aws/s3/model/ListObjectsV2Request.h>
 #include <aws/s3/model/ListPartsRequest.h>
 
 #pragma GCC diagnostic pop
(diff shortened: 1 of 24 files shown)


src/file_proxy_s3.cc
```diff
@@ -276,6 +276,32 @@ s3_delete_object(const char* file_path)
...
+		if (res.GetIsTruncated()) {
+			const Aws::String& ct = res.GetNextContinuationToken();
+			req.SetContinuationToken(ct);
+		}
+
+		for (const Aws::S3::Model::Object& object : res.GetContents()) {
+			target.push_back(object);
+		}
+	}
+	while (outcome.GetResult().GetIsTruncated());
```
jdogmcsteezy, line 300: might not be possible, but maybe save the result of `outcome.GetResult().GetIsTruncated()` since you use it in the loop.  I am not sure if those function calls are idempotent or not

src/file_proxy_s3.cc
```diff
@@ -297,17 +323,17 @@ s3_delete_directory(const char* dir_path)
...
+	Aws::S3::Model::ListObjectsV2Request req;
 	req.SetBucket(path.GetBucket());
 	req.SetPrefix(path.GetKey());
 
-	Aws::S3::Model::ListObjectsOutcome res = client.ListObjects(req);
-	if (!res.IsSuccess()) {
-		err("%s", res.GetError().GetMessage().c_str());
+	Aws::Vector<Aws::S3::Model::Object> res;
+	bool success = ListAllObjects(client, req, res);
+	if (!success) {
```
jdogmcsteezy: Just a suggestion. Do we need the variable for later?  It seems more inline with Clayton's style to use the func as a condition. Your call 
```suggestion
	if (!ListAllObjects(client, req, res)) {
```

jdogmcsteezy (approved):
Looks fine to me.  Just some suggestions.  

I think you should edit the commit and summary before merging and be more specific about what you are fixing.  Backups work fine.  Restores do not

src/file_proxy_s3.cc
dwelch-spike: The only functions called on `res` are getters that are idempotent.

src/file_proxy_s3.cc
dwelch-spike: Agreed, changed all instances.

PR #48, opened 2023-01-10
Use openssl 3.0, update aws sdk

dwelch-spike:


dwelch-spike:
Canceled the test runs for the readme changes. These changes are passing tests.

src/s3_api.cc
```diff
@@ -263,9 +263,8 @@ S3API::_init_api(S3API& s3_api)
 	Aws::InitAPI(s3_api.options);
 
 	Aws::Client::ClientConfiguration conf;
-	if (s3_api.region.empty() && s3_api.endpoint.empty()) {
-		err("S3 API must be enabled by specifying a region if no endpoint "
-				"override is given.");
+	if (s3_api.region.empty()) {
+		err("S3 API must be enabled by specifying a region.");
```
jdogmcsteezy: I guess a region is always required now and before providing a URL was enough? Is that considered breaking? Maybe there is a way around this

jdogmcsteezy: maybe it was a bug?

jdogmcsteezy (approved):
LGTM, but a question that might affect the next release version

src/s3_api.cc
dwelch-spike: I couldn't find any mention of this when I upgraded but I see something that reverts it in aws-cpp-sdk 1.10.55 https://github.com/aws/aws-sdk-cpp/pull/2295 . I upgraded to that and confirmed that no region works again.

PR #49, opened 2023-01-19
TOOLS-1971 add example conf document default tls protocol

dwelch-spike:


jdogmcsteezy (approved):
LGTM. Just a thought though.  What if we removed astools.conf from all the tools and had a link in the README to a single astools.conf file.  That way, in the future, we do not need to make changes to every submodule

dwelch-spike:
> LGTM. Just a thought though. What if we removed astools.conf from all the tools and had a link in the README to a single astools.conf file. That way, in the future, we do not need to make changes to every submodule

This makes sense to me. Some of the tools might require a local copy for testing but I'll make a ticket for this.

PR #50, opened 2023-01-24
Auto draft release

dwelch-spike:
The diff is larger than it should be because I have included https://github.com/aerospike/aerospike-tools-backup/pull/48 to make sure the mac build passes. Please review that first.

Relates to TOOLS-2305 in order to record snyk reports with releases (not done here).

Example of what the draft will look like. I'm hoping more will show up in the diff section of an actual release, that might require some tweaking.
image

I'm hoping to spread this to other repos.
Will squash merge and discard all the garbage messages.

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -1,0 +1,43 @@
+---
+name: "Draft Release"
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  draft-release-notes:
+    needs: build-asbackup-mac
+    runs-on: ubuntu-latest
+    name: Draft Release
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Get Latest Tag
+      run: |
+        LATEST_TAG=$(git describe --tags `git rev-list --tags --max-count=1`)
+        echo "LATEST_TAG=$LATEST_TAG" >> $GITHUB_ENV
+    # below steps are for downloading release artifacts
+    - name: Download Asbackup Mac Artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: darwin-X64-macOS-11-asbackup-static
+        path: artifacts
+    - name: Download Asrestore Mac Artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: darwin-X64-macOS-11-asrestore-static
+        path: artifacts
+    # finally create the release and upload artifacts
+    - name: Upload Artifacts to Release Draft
+      uses: "softprops/action-gh-release@v1"
+      with:
+        draft: true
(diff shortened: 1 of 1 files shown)


dwelch-spike:
Canceled the last run because it had duplicates. See the passed checks are identical. I think this happened because I deleted the test tag.

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,43 @@
...
+    needs: build-asbackup-mac
+    runs-on: ubuntu-latest
+    name: Draft Release
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Get Latest Tag
+      run: |
+        LATEST_TAG=$(git describe --tags `git rev-list --tags --max-count=1`)
```
jdogmcsteezy: curious what `rev-list` is there for?

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,43 @@
...
+        echo "LATEST_TAG=$LATEST_TAG" >> $GITHUB_ENV
+    # below steps are for downloading release artifacts
+    - name: Download Asbackup Mac Artifact
```
jdogmcsteezy: Not sure we should upload any artifacts unless we provide them all. Seems weird to me.  Plus, these are unnotorized

jdogmcsteezy: all meaning "mac, deb, rpm"

jdogmcsteezy (commented):
I am confused about what this action is actually doing.  It drafts a release but does not automate anything else.  Not sure this really speeds any process up but adds one more workflow to this project that has a lot of unused workflows already

.github/workflows/draft-release.yml
dwelch-spike: its function here is to list the most recent commit.

.github/workflows/draft-release.yml
dwelch-spike: will remove for now.

dwelch-spike:
tested
image

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,19 @@
...
+  draft-release-notes:
+    runs-on: ubuntu-latest
+    name: Draft Release
+    steps:
+    - uses: actions/checkout@v3
+    - name: Upload Artifacts to Release Draft
+      uses: "softprops/action-gh-release@v1"
+      with:
+        draft: true
+        generate_release_notes: true
```
jdogmcsteezy:
```suggestion
        generate_release_notes: true
        
```

jdogmcsteezy (approved):
LGTM

PR #51, opened 2023-02-07
TOOLS-2327 TOOLS-2371 Add --directory-list, --parent-directory options to support restore from multiple directories.

dwelch-spike:


src/restore.c
```diff
@@ -153,13 +153,58 @@ restore_main(int32_t argc, char **argv)
...
+	// restoring from multiple directories
+	if (conf.directory_list != NULL) {
+
+		uint32_t dir_count = 0;
+
+		char *dir_clone = safe_strdup(conf.directory_list);
+		split_string(dir_clone, ',', false, &directories);
+		dir_count = directories.size;
+
+		for (uint32_t i = 0; i < dir_count; i++) {
```
```suggestion
		char *dir_clone = safe_strdup(conf.directory_list);
		split_string(dir_clone, ',', false, &directories);
		uint32_t dir_count = directories.size;

		for (uint32_t i = 0; i < dir_count; i++) {
```
Do we need dir_count anyway?

src/restore.c
```diff
@@ -153,13 +153,58 @@ restore_main(int32_t argc, char **argv)
...
+			char *dir = as_vector_get_ptr(&directories, i);
+
+			if (conf.parent_directory) {
+				size_t path_size = strlen(conf.parent_directory) + strlen(dir) + 1;
+				char *tmp_dir = dir;
+				dir = cf_malloc(path_size);
+				snprintf(dir, path_size, "%s%s", conf.parent_directory, tmp_dir);
```
jdogmcsteezy: What happens when parent directory does not end with `/`? I feel `/` should not be a requirement

src/restore_config.c
```diff
@@ -603,6 +618,16 @@ restore_config_init(int argc, char* argv[], restore_config_t* conf)
...
+	if (conf->directory && conf->directory_list) {
+		err("Invalid options: --directory and --directory-list are mutually exclusive.");
+		return RESTORE_CONFIG_INIT_FAILURE;
+	}
+
+	if (conf->input_file && conf->directory_list) {
+		err("Invalid options: --input-file and --directory-list are mutually exclusive.");
+		return RESTORE_CONFIG_INIT_FAILURE;
+	}
+
```
jdogmcsteezy, line 630: What about `--input-file` and `--directory`? Is that invalid?

src/restore_config.c
```diff
@@ -957,10 +992,18 @@ usage(const char *name)
 	fprintf(stdout, "                      Used to restore to a different namespace.\n");
 	fprintf(stdout, "  -d, --directory <directory>\n");
 	fprintf(stdout, "                      The directory that holds the backup files. Required, \n");
-	fprintf(stdout, "                      unless -i is used.\n");
+	fprintf(stdout, "                      unless -i or --directory-list is used.\n");
 	fprintf(stdout, "  -i, --input-file <file>\n");
 	fprintf(stdout, "                      Restore from a single backup file. Use - for stdin.\n");
-	fprintf(stdout, "                      Required, unless -d is used.\n");
+	fprintf(stdout, "                      Required, unless -d or --directory-list is used.\n");
```
jdogmcsteezy:
```suggestion
	fprintf(stdout, "                      Required, unless --directory or --directory-list is used.\n");
```

src/restore_config.c
```diff
@@ -957,10 +992,18 @@ usage(const char *name)
...
+	fprintf(stdout, "      --directory-list <path 1>[,<path2[,...]]\n");
+	fprintf(stdout, "                      A comma seperated list of paths to directories that hold the backup files. Required, \n");
+	fprintf(stdout, "                      unless -i or -d is used. The paths may not contain commas.\n");
```
jdogmcsteezy, line 1001: I don't think commas are allowed in paths anyway? Not sure if that needs to be said

jdogmcsteezy (changes requested):
Mostly some questions and suggestions

src/restore_config.c
dwelch-spike: Yes, there is a check for that on line 602.

src/restore_config.c
dwelch-spike: I've never seen it but they are allowed.

src/restore.c
dwelch-spike: If the paths in --directory-list start with '/' it will work fine, otherwise the path parsing fails. Concatenating the strings gives users the most freedom.

src/restore.c
dwelch-spike: No, removed it.

src/restore.c
jdogmcsteezy: Yeah, but most applications don't care whether you add or remove the trailing `/`. I think it would be a nicer user experience if they can just provide them without having to know whether to add the trailing `/`. All I am saying is, I would be annoyed if the problem was because I needed to add an extra `/`.  I would think there is a simple `path_join(...)` but probably not because this is c :). As long as the error message is clear I am good with it

src/restore.c
```diff
@@ -153,13 +153,55 @@ restore_main(int32_t argc, char **argv)
...
+			}
+
+			if (!get_backup_files(dir, &status.file_vec)) {
+				err("Error while getting backup files from directory_list entry: %u", i);
```
jdogmcsteezy: Whether you pick to make the `/` optional or not this should probably change so they know if they formatted it wrong. This should probably go for other error messages as well where asbackup fails because of an incorrect path or something
```suggestion
				err("Error while getting backup files from directory_list entry: %s", dir);
```

src/restore_config.c
jdogmcsteezy: Maybe we could move that check closer to the other mutual exclusion checks. Should these checks also use `!=  NULL` like the others?

jdogmcsteezy (changes requested):

src/restore.c
dwelch-spike: Since the option is called "parent-directory" I think this is reasonable to do. It was easy enough so added it.

src/restore_config.c
dwelch-spike: makes sense, done

jdogmcsteezy (approved):
LGTM

PR #56, opened 2023-03-30
perf: TOOLS-2450 don't get backup file sizes individually

dwelch-spike:


src/file_proxy.c
```diff
@@ -1029,33 +1004,32 @@ disk_space_remaining(const char *dir)
...
 
 	ver("Listing backup files in %s", dir_path);
 
 	switch (file_proxy_type) {
 		case FILE_PROXY_TYPE_S3:
-			res = s3_get_backup_files(dir_path, file_vec);
+			total_file_size = s3_get_backup_files(dir_path, file_vec);
 			break;
 
 		case FILE_PROXY_TYPE_LOCAL:
```
jdogmcsteezy, line 1022: Maybe it would be good to create another function to mirror `s3_get_backup_files` like `local_get_backup_files`

src/restore.c
```diff
@@ -220,17 +223,8 @@ restore_main(int32_t argc, char **argv)
...
-				off_t size = get_file_size(path);
-				if (size == -1) {
-					err("Failed to get the size of file %s", path);
-					goto cleanup5;
-				}
-
-				status.estimated_bytes += size;
-			}
+			status.estimated_bytes = total_file_size;
+			ver("Estimated total backup file size %lli bytes", status.estimated_bytes);
```
```suggestion
			ver("Estimated total backup file size: %lli bytes", status.estimated_bytes);
```

jdogmcsteezy (approved):
LGTM just a couple suggestions

src/file_proxy.c
dwelch-spike: I noticed that the local case doesn't have an abstraction. I might add this in the future.

PR #58, opened 2023-05-22
feat: TOOLS-2514 add --s3-connect-timeout

dwelch-spike:
Also add missing s3 config unit tests.


src/conf.c
```diff
@@ -847,6 +847,14 @@ config_backup(toml_table_t *conftab, backup_config_t *c, const char *instance,
 				status = false;
 			}
 
+		} else if (! strcasecmp("s3-connect-timeout", name)) {
```
jdogmcsteezy, line 850: Only nit picky thing I can say it to keep the flags in the same order in both backup and restore

src/restore_config.c
```diff
@@ -1125,7 +1136,10 @@ usage(const char *name)
 	fprintf(stdout, "                       - Info\n");
 	fprintf(stdout, "                       - Debug\n");
 	fprintf(stdout, "                       - Trace\n");
-	fprintf(stdout, "                      The default is Fatal.\n\n");
+	fprintf(stdout, "                      The default is Fatal.\n");
+	fprintf(stdout, "      --s3-connect-timeout <ms>\n");
+	fprintf(stdout, "                      The internal aws s3 client's connection timeout in milliseconds.\n");
+	fprintf(stdout, "                      Corresponds to the connectTimeoutMS aws-sdk-cpp client configuration.\n\n");
```
jdogmcsteezy:
```suggestion
	fprintf(stdout, "                      The default is Fatal.\n");
	fprintf(stdout, "      --s3-connect-timeout <ms>\n");
	fprintf(stdout, "                      The AWS S3 SDK's connection timeout in milliseconds.\n");
```
The above description uses uppercase and calls it the SDK. This is fine. The client or SDK does not matter to the user.  You don't need to tell them more about what it corresponds to in the actual SDK. If your description does not provide enough info, add it here. Don't make them go looking. They might not be as savvy as we devs

test/integration/test_s3.py
```diff
@@ -165,7 +165,7 @@ def test_s3_backup():
 	do_s3_backup(0)
 
 def test_s3_backup_multiple_files():
```
jdogmcsteezy, line 167: Any fail test? Maybe that is not possible? Just looking for confirmation it is working as intended

jdogmcsteezy (approved):
LGTM, some suggestions that I am not attached to. Main thing is some time of test to verify it is working

src/conf.c
dwelch-spike: They are but restore does not have s3_max_async_downloads so it looks different

src/restore_config.c
dwelch-spike: I know most users won't know it but this is specific to the Client class in the SDK, that's why I call it "client".
I used the capitalization suggestions, I'd like to keep the mention of the equivalent config items because it lets people look them up for more information.

test/integration/test_s3.py
dwelch-spike: Adding new s3 tests to these isn't easy, it's very structured, but negative tests are something I'll look into.

PR #60, opened 2023-06-29
fix: Tools-2569 correctly parse --s3-bucket

dwelch-spike:
This is scheduled for tools 8.6.


dwelch-spike:
We are opting to remove --s3-bucket instead.

PR #61, opened 2023-07-05
fix: TOOLS-2569 remove --s3-bucket

dwelch-spike:


dwelch-spike (commented):
LGTM (reviewed with Jesse)

PR #62, opened 2023-07-25
fix: tools-2611 deadlock, only increment async_downloads when there are more par…

dwelch-spike:
…ts to download

### DIFF
--- a/src/download_manager.cc
+++ b/src/download_manager.cc
@@ -372,7 +372,6 @@ GroupDownloadManager::StartNextPart()
 		return true;
 	}
 
-	async_downloads++;
 	DownloadManager* next = dms.front();
 
 	// place next back on the end of the queue if it has more parts that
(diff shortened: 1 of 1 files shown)


jdogmcsteezy (commented):
Can you add some explanation for the change?

dwelch-spike:
> Can you add some explanation for the change?

`InitiateDownloadNextPart` checks if the file being downloaded has more parts to download, and starts the next one if it does. `async_downloads` should only be incremented if another download is actually started. Before, it was being incremented even if `InitiateDownloadNextPart()` didn't start the next part. By moving it under the successful `InitiateDownloadNextPart()` case, accounting is done correctly.

These extra increments weren't cleared in the case of the bug and would add up 1 per backup file opened. After the backup thread opened --s3-max-downloads backup files, it would deadlock, because async_downloads == --s3-max-downloads.

jdogmcsteezy:
I'm not sure I completely understand the mechanics here but I will approve anyway.  I am confused about the case where `StartNextPart` is called but `InitiateDownloadNextPart()` does not return true

jdogmcsteezy (approved):

dwelch-spike:
`StartNextPart` is called anytime a part is finished downloading. If the finished part was the last part, `StartNextPart` calls `InitiateDownloadNextPart` which checks to see if any more downloads need to be done, and will return false without starting a download if all parts have finished.

PR #63, opened 2023-07-31
fix: tools-2624 correctly set bin-list for each thread

dwelch-spike:
split_string() was being called on the same bin_list string from each thread, mangling it and providing a different bin_list for each thread.

### DIFF
--- a/src/backup.c
+++ b/src/backup.c
@@ -2255,12 +2255,12 @@ init_scan_bins(char *bin_list, as_scan *scan)
 	as_vector bin_vec;
 	as_vector_inita(&bin_vec, sizeof (void *), 25);
 
-	if (bin_list[0] == 0) {
+	if (clone[0] == 0) {
 		err("Empty bin list");
 		goto cleanup1;
 	}
 
-	split_string(bin_list, ',', true, &bin_vec);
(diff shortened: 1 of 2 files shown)


src/backup.c
```diff
@@ -2255,12 +2255,12 @@ init_scan_bins(char *bin_list, as_scan *scan)
...
 	as_vector_inita(&bin_vec, sizeof (void *), 25);
 
-	if (bin_list[0] == 0) {
+	if (clone[0] == 0) {
 		err("Empty bin list");
 		goto cleanup1;
 	}
 
-	split_string(bin_list, ',', true, &bin_vec);
+	split_string(clone, ',', true, &bin_vec);
```
jdogmcsteezy, line 2263: Is the problem that split_string mangles the bin_list or that the threads mangle the bin_list because pointers to bin_list are stored in the bin_vec?

jdogmcsteezy (approved):
LGTM

src/backup.c
dwelch-spike: Kind of both, each thread has a pointer to the same list, and split_string mangles them. So each thread ends up with a different view of the string when they parse it.

PR #65, opened 2023-08-08
fix: tools-2630 correctly close/open backup files in error states

dwelch-spike:


dwelch-spike:
This PR fixes a couple of issues that occur when asbackup fails to open or close a backup file.

1. When a backup file fails to open, free it and don't add it to the file queue to be placed in the state file. This was causing bad reads when the state file was written to disk and tried to read a partially opened backup file.
2. If a backup file fails to close, only push it to the file queue once. This was causing double free errors where files would fail to close in the scan_callback function, then again in the backup_work function.
3. Do not attempt to close NULL backup files. This is mostly used as a way to avoid closing backup files twice.

Also added some tests that overload the libc fopen and fread calls in order to simulate file closing failures.

src/backup.c
```diff
@@ -1136,7 +1145,7 @@ close_file(io_write_proxy_t *fd)
 
 	int res = io_proxy_close2(fd, FILE_PROXY_EOF);
 	if (res != 0) {
-		err("Error while closing backup file");
+		err("Error while closing backup io proxy");
```
jdogmcsteezy, line 1148: Isn't this almost always a file unless backing up to S3?

src/backup.c
```diff
@@ -1343,6 +1362,9 @@ open_dir_file(backup_job_context_t *bjc)
 		if (update_file_pos(bjc->fd, &bjc->byte_count_file, &bjc->byte_count_job,
 					&bjc->status->byte_count_total) < 0) {
 			pthread_mutex_unlock(&bjc->status->dir_file_init_mutex);
+			cf_free(bjc->fd);
+			bjc->fd = NULL;
+			err("New directory file %s, has less than 0 bytes written", file_path);
```
jdogmcsteezy: What is less than 0 bytes? A nibble? A bit?

jdogmcsteezy (commented):
Looks minimal to me. You clean up the fd earlier rather than at the end. I just have a couple of output/doc questions

src/backup.c
dwelch-spike: Since this function deals in io proxies, I changed the err text here. This error message was duplicated throughout  the code, I wanted to make more of them unique for easier debugging.

src/backup.c
dwelch-spike: yeah that was confusing. changed

jdogmcsteezy: I would have changed to ordering to "Failed to get file position from new directory file %s"

jdogmcsteezy (approved):
LGTM

PR #66, opened 2023-08-18
build: tools-2643 add make shared, builds asbackup as a dynamic lib

dwelch-spike:


dwelch-spike:
Closing untill backup_run is exposed

Makefile
```diff
@@ -320,11 +332,22 @@ TEST_DEPS := $(sort $(TEST_DEPS))
 .PHONY: all
 all: $(BINS)
 
+.PHONY: set_dynamic_options
+set_dynamic_options: $(TOML)
+	$(eval DYNAMIC_OPTIONS = -fPIC)
```
jdogmcsteezy: What is this for? Would I ever call `make set_dynamic_options`?

src/restore.c
```diff
@@ -38,11 +38,11 @@
 static restore_config_t* g_conf;
 static restore_status_t* g_status;
 
-
 //==========================================================
 // Forward Declarations.
 //
 
+static restore_status_t* start_restore(restore_config_t *conf);
```
jdogmcsteezy, line 45: Is it normal to have a forward declaration of your new restore_run() function?

src/restore.c
```diff
@@ -333,30 +393,30 @@ restore_main(int32_t argc, char **argv)
 		}
 	}
 
-	if (!conf.validate && !batch_uploader_await(&status.batch_uploader)) {
+	if (!conf->validate && !batch_uploader_await(&status->batch_uploader)) {
 		res = EXIT_FAILURE;
 	}
 
 	// NOTE this is here to support the --indexes-last option
```
jdogmcsteezy, line 403: Tangent. I wonder why --indexes-last is even needed. I would think indexes-last should always be true

src/restore.c
```diff
@@ -388,23 +448,17 @@ restore_main(int32_t argc, char **argv)
...
+	}
 
 cleanup1:
-	restore_config_destroy(&conf);
 
-	file_proxy_cloud_shutdown();
-
-	ver("Exiting with status code %d", res);
-
-	return res;
```
jdogmcsteezy, line 400: I assume the caller now handles this stuff. It's a bit difficult to read the diff because it is mostly code moved around

src/restore.c
```diff
@@ -81,61 +81,121 @@ restore_main(int32_t argc, char **argv)
...
+ * Runs a restore job with the given configuration. This method is not thread
+ * safe and should not be called multiple times in parallel, as it uses global
+ * variables to handle signal interruption.
+ *
+ * Returns the restore_status struct used during the run which must be freed by the
+ * caller.
+ * Only free the return value if it is != RUN_RESTORE_FAILURE
+ */
+restore_status_t*
+start_restore(restore_config_t *conf)
```
```suggestion
static restore_status_t*
start_restore(restore_config_t *conf)
```

jdogmcsteezy (changes requested):

Makefile
dwelch-spike: I used it as a way to require this statement as a prerequisite to BACKUP_DYNAMIC and RESTORE_DYNAMIC when running `make shared`, you wouldn't run `make set_dynamic_options` as a user.

src/restore.c
dwelch-spike: start_restore is now a local helper, it's pattern in these files to forward declare local helpers and is required if they aren't listed in any imported header files. I should add `static` to the definition though, will do.

src/restore.c
dwelch-spike: Why? Do you get an error if nothing exists for them to index?

src/restore.c
dwelch-spike: If the user of the shared lib uses `restore_run` it handles the cloud shutdown but the caller needs to destroy the conf. It's outlined in the docstring for `restore_run`

Makefile
dwelch-spike: Added a comment

jdogmcsteezy: seems weird since it isn't really building anything. Would it not always run when running `make shared`?

src/restore.c
jdogmcsteezy: When you write a record and an index already exists it takes some time for the record to be indexed by the sindex. If you write the record first then create the sindex all existing records are immediately indexable

jdogmcsteezy (approved):

src/restore.c
dwelch-spike: Hmmm, that does seem like it should be the default. I'm going to make a ticket for it.

dwelch-spike: Made TOOLS-2675 to track this

PR #67, opened 2023-09-13
Tools-2667

dwelch-spike:
set s3 config items in the shared lib run functions


jdogmcsteezy (approved):
This change seems simple enough. I am only looking at the 2 `set_s3_configs()` functions which were moved out of the `src/*_config.c` files. The rest of the changes appear to be apart of #66

PR #68, opened 2023-09-19
tools-2680 tools-2681 backup service changes

dwelch-spike:
PR to run tests

### DIFF
--- a/src/s3_api.cc
+++ b/src/s3_api.cc
@@ -52,9 +52,11 @@ S3API::S3API() : initialized(false),
 bool
 S3API::TryInitialize()
 {
-	std::call_once(init_once, _init_api, std::ref(*this));
+	if(!initialized) {
+		_init_api(std::ref(*this));
+	}
 
-	return this->initialized;
+	return initialized;
 }
 
 void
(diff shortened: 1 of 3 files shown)


dwelch-spike:
@reugn I moved your commits here. I also changed the s3_init function to lock. These changes look good to me. If they look good to you and the tests pass I'll merge them. Please use this branch for development in the mean time.

src/s3_api.cc
```diff
@@ -52,27 +52,38 @@ S3API::S3API() : initialized(false),
 bool
 S3API::TryInitialize()
 {
-	std::call_once(init_once, _init_api, std::ref(*this));
+	if(initialized) {
+		return initialized;
+	}
```
reugn: It's better not return here without synchronization, as the value might be read from the cache.
I would rather use `std::atomic<bool>` for the `initialized` field. That will also remove the lock from the `IsInitialized` method.

dwelch-spike: Agreed, this should be under lock. changed that.

I think relying on `initialized` as an atomic in `IsInitialized` could lead to a data race. For example, if one thread has called `TryInitialize` and is executing `_init_api` `initialized` might not be set yet, causing `IsInitialized` to return false even though the s3api is in the process of being initialized. This could cause problems like `SetLogLevel` setting the s3 log level after it has been initialized so that the new value won't take effect.

reugn (approved):

PR #72, opened 2023-09-29
feat: Tools 2539 secret agent support

dwelch-spike:
Much of this work is in the new Python test files, which I think don't need to be reviewed.

The bulk of the C changes are in conf.c, backup_conf.c and restore_conf.c.
They use the secret agent to pull secrets out of configs/options with the format secrets:<resource_name>:<secret_name>. 
A few gotchas ...

- The secret client returns a heap allocated result on success so these files have to manage that memory. 
- The toml parsing function toml_rtos allocates heap memory for its return value so that needs to be managed.


.github/workflows/backup-build.yml
```diff
@@ -15,6 +15,9 @@ jobs:
       EVENT_LIB: libuv
       ZSTD_VERSION: 1.4.9
     steps:
+    - uses: actions/setup-go@v4
+      with:
+        go-version: "^1.19.0"
```
jdogmcsteezy: Make this variable come from `go.mod`. The unique-data agent does this. This is an easy thing to forget to update. Aren't we using 1.20 to build using bob?

.github/workflows/backup-build.yml
```diff
@@ -134,6 +141,31 @@ jobs:
...
+    - name: Test secret agent
+      run: |
+        make run_test_secrets
+      working-directory: main
+    - name: Test secrets
+      run: |
+        make run_test_secret_agent
+      working-directory: main
+    # - name: Enable core directories
+    #   if: always()
```
jdogmcsteezy: I think that is an `if: fails()` or something similar if you want that

src/backup_config.c
```diff
@@ -154,6 +154,11 @@ backup_config_init(int argc, char* argv[], backup_config_t* conf)
 		{ "s3-max-async-uploads", required_argument, NULL, COMMAND_OPT_S3_MAX_ASYNC_UPLOADS },
 		{ "s3-log-level", required_argument, NULL, COMMAND_OPT_S3_LOG_LEVEL },
 		{ "s3-connect-timeout", required_argument, NULL, COMMAND_OPT_S3_CONNECT_TIMEOUT },
+
+		{ "sa-address", required_argument, NULL, COMMAND_SA_ADDRESS },
```
jdogmcsteezy, line 158: It is odd to me that every arg is a "required_argument"

src/backup_config.c
```diff
@@ -388,10 +499,19 @@ backup_config_init(int argc, char* argv[], backup_config_t* conf)
 				err("Cannot specify both encryption-key-file and encryption-key-env");
 				return BACKUP_CONFIG_INIT_FAILURE;
 			}
+
 			conf->pkey = (encryption_key_t*) cf_malloc(sizeof(encryption_key_t));
-			if (io_proxy_read_private_key_file(optarg, conf->pkey) != 0) {
-				return BACKUP_CONFIG_INIT_FAILURE;
+			if (arg_is_secret) {
+				if(io_proxy_read_private_key(optarg, conf->pkey) != 0) {
```
jdogmcsteezy: Why are we going through the io_proxy to read a secret? Isn't this coming through the secret agent?

jdogmcsteezy: I think I am confused by what this pkey is for. I thought it was tls related but it looks like it is the pkey for encrypting the backup files

src/backup_config.c
```diff
@@ -1336,6 +1505,34 @@ usage(const char *name)
 	fprintf(stdout, " --config-file <path>\n");
 	fprintf(stdout, "                      Read this file after default configuration file.\n");
 	fprintf(stdout, " --only-config-file <path>\n");
-	fprintf(stdout, "                      Read only this configuration file.\n");
+	fprintf(stdout, "                      Read only this configuration file.\n\n");
+
+	fprintf(stdout, "[secret-agent]\n");
+	fprintf(stdout, " Options pertaining to the Aerospike secret agent https://docs.aerospike.com/tools/secret-agent.\n");
```
jdogmcsteezy, line 1511:
```suggestion
	fprintf(stdout, " Options for connecting to the Aerospike secret agent https://docs.aerospike.com/tools/secret-agent.\n");
```

src/backup_config.c
```diff
@@ -1336,6 +1505,34 @@ usage(const char *name)
 	fprintf(stdout, " --config-file <path>\n");
 	fprintf(stdout, "                      Read this file after default configuration file.\n");
 	fprintf(stdout, " --only-config-file <path>\n");
-	fprintf(stdout, "                      Read only this configuration file.\n");
+	fprintf(stdout, "                      Read only this configuration file.\n\n");
+
+	fprintf(stdout, "[secret-agent]\n");
```
jdogmcsteezy, line 1510: Is this is a global section in the config file that can be used by other tools in the future?

src/conf.c
```diff
@@ -28,6 +28,7 @@
 #include <dirent.h>
 #include <sys/stat.h>
 
+#include <sc_client.h>
```
jdogmcsteezy: This name is confusing since you refer. to the secret agent as `sa` in a lot of the code. `sc` does not make me think secret-agent

src/conf.c
```diff
@@ -266,63 +438,90 @@ config_restore_cluster(toml_table_t *conftab, restore_config_t *c, const char *i
...
+			return false;
+		}
 
+		bool arg_is_secret = false;
+		char *override = NULL;
+		if (!get_secret_rtoml(sc, config_value, &override, &arg_is_secret)) {
+			return false;
+		}
+
+		bool status;
```
jdogmcsteezy, line 454: Would it be possible to set config_value = override if override != NULL rather than changing config_str to fit this specific case?

src/conf.c
```diff
@@ -197,56 +251,174 @@ config_str(toml_table_t *curtab, const char *name, void *ptr)
...
 	return false;
 }
 
 static bool
-config_bool(toml_table_t *curtab, const char *name, void *ptr)
+config_bool(const char *raw_val, void *ptr, const char *override)
 {
-	const char *value = toml_raw_in(curtab, name);
-	if (! value) {
+	if (override != NULL) {
```
jdogmcsteezy, line 306: I think it would be more readable if these functions were reverted rather than turned into and if/else helper

jdogmcsteezy: If override != null then set raw_val = override from the caller

src/conf.c
```diff
@@ -197,56 +251,174 @@ config_str(toml_table_t *curtab, const char *name, void *ptr)
...
+
+	char secret_agent[256] = {"secret-agent"};
+	if (instance) {
+		snprintf(secret_agent, 255, "secret-agent_%s", instance);
+		// override if it exists otherwise use
+		// default section
+		if (toml_table_in(conftab, secret_agent)) {
+			curtab = toml_table_in(conftab, secret_agent);
+		}
+	}
```
jdogmcsteezy: why no if else?

src/conf.c
```diff
@@ -197,56 +251,174 @@ config_str(toml_table_t *curtab, const char *name, void *ptr)
...
 	return false;
 }
 
 static bool
-config_restore_cluster(toml_table_t *conftab, restore_config_t *c, const char *instance,
+config_secret_agent(toml_table_t *conftab, sc_cfg *c, const char *instance,
 		char errbuf[])
+{
+	// Defaults to "secret-agent" section in case present.
+	toml_table_t *curtab = toml_table_in(conftab, "secret-agent");
```
jdogmcsteezy: curtab? Looks like the precedent was set with conftab but we use snake_case everywhere else and curtab does not make me think current_table. Makes me think current_tab

src/conf.c
```diff
@@ -197,56 +251,174 @@ config_str(toml_table_t *curtab, const char *name, void *ptr)
...
+
+	if (! curtab) {
+		return true;
+	}
+
+	const char *name;
+	bool used_sa_port_arg = false;
+
+	for (uint8_t k = 0; 0 != (name = toml_key_in(curtab, k)); k++) {
+		char* config_value = (char*) toml_raw_in(curtab, name);
```
jdogmcsteezy: How about `curval`? :)

src/io_proxy.c
```diff
@@ -145,6 +145,38 @@ io_proxy_read_private_key_file(const char* pkey_file_path,
 	return 0;
 }
 
+int
+io_proxy_read_private_key(char* pkey_data,
```
jdogmcsteezy: I am still confused as to why reading a private_key is handled by the IO proxy

src/utils.c
```diff
@@ -2011,3 +2068,127 @@ tls_config_clone(as_config_tls* clone, const as_config_tls* src)
...
+		return NULL;
+	}
+
+	buf[fsize] = 0;
+
+	return buf;
+}
+
+int
+get_and_set_secret_arg(sc_client* sc, char* path, char** res, bool* is_secret) {
```
jdogmcsteezy: Why is it called `get_and_set`? It looks like it only gets?

jdogmcsteezy (changes requested):
Other than some naming/style choices, my main concern is that it does not appear (I'm only looking at diffs) that adding `env:`, `env-64` and `file` support will be isolated/easy in the future. I could be wrong though, I can only gather so much from the diffs

src/conf.c
dwelch-spike: If I made config_str() aware of secrets with a flag, and it handled quoting secrets, I could do this. I don't want to right now though. I'll think about it.

src/conf.c
dwelch-spike: You mean why add `override` to `config_str`? toml raw strings are always quoted, so I would have to change `config_str` to handle that, I opted to use a blanket override instead. I didn't want to modify the secret data or have to realloc or anything.

src/backup_config.c
dwelch-spike: Yes, it's like the "cluster" section.

src/backup_config.c
dwelch-spike: Matched the io_proxy_read_private_key_file() function here. These functions read the private key that io_proxies use for encryption/decryption. They don't actually use io_proxies. When this is called, `optarg`, contains the secret value for the private key and io_proxy_read_private_key() parses it.

src/backup_config.c
dwelch-spike: This means that the option "sa-address" requires an argument.

src/conf.c
dwelch-spike: What do you mean? This is a pattern taken from other places in this file, I'm trying to keep things consistent.

src/io_proxy.c
dwelch-spike: The names are confusing. I followed the existing pattern for reading the encryption key. The value is used by the io_proxy but the functions do not use io_proxies. I think the pattern for the io_proxy file is to prefix everything with that name. 

I'll look into moving them into the utils file.

src/utils.c
dwelch-spike: renamed

src/conf.c
dwelch-spike: I think changing the names and making these snake case makes sense. Changed.

src/conf.c
dwelch-spike: This might change soon, I'll probably rename the secrets-client-c lib

.github/workflows/backup-build.yml
dwelch-spike: thanks, changed for future use.

.github/workflows/backup-build.yml
dwelch-spike: This was just for building the secret agent for tests. Now that it has a docker container I don't need it. Removed.

dwelch-spike:
> Other than some naming/style choices, my main concern is that it does not appear (I'm only looking at diffs) that adding `env:`, `env-64` and `file` support will be isolated/easy in the future. I could be wrong though, I can only gather so much from the diffs

Yeah this isn't the best. I don't think it would be too bad though.

I have a refactor in mind to make it cleaner. I think the config parsing should be a pipeline, and each stage can handle a source. For example, `file` stage, then `command line args` then`env` then `env-64` stage etc. If we have to support another set of config sources I'll try to make this happen, I didn't do it now in order to get this done faster.

When it comes time to support secrets in other tools I think we should try to do it that way, atleast for the C tools.

dwelch-spike:
> > Other than some naming/style choices, my main concern is that it does not appear (I'm only looking at diffs) that adding `env:`, `env-64` and `file` support will be isolated/easy in the future. I could be wrong though, I can only gather so much from the diffs
> 
> Yeah this isn't the best. I don't think it would be too bad though.
> 
> I have a refactor in mind to make it cleaner. I think the config parsing should be a pipeline, and each stage can handle a source. For example, `file` stage, then `command line args` then`env` then `env-64` stage etc. If we have to support another set of config sources I'll try to make this happen, I didn't do it now in order to get this done faster.
> 
> When it comes time to support secrets in other tools I think we should try to do it that way, atleast for the C tools.

Spoke to Jesse,

Idea is use is_secret as a more generic marker for if an option comes from a source other than the command line or config file. Maybe write a parsing function that wraps get_secret and handles checking the option prefix for "secret:", "env:", etc.

src/io_proxy.c
dwelch-spike: moved them to utils and renamed them.

src/conf.c
dwelch-spike: keeping this

src/conf.c
dwelch-spike: keeping this

src/conf.c
dwelch-spike: keeping this

src/conf.c
dwelch-spike: renamed to sa_client. Secret client renamed to secret-agent-client-c

src/backup_config.c
dwelch-spike: This might grow to include options that aren't connection related. I want to keep the usage information general.

src/backup_config.c
dwelch-spike: renamed and moved to utils

jdogmcsteezy (approved):
As long as the tests pass LGTM

PR #74, opened 2023-10-20
build: tools-2539 secret agent client use jansson submodule

dwelch-spike:


dwelch-spike:
Opted to keep jansson as an external dependency

PR #75, opened 2023-11-14
Tools 2767 support sindex on blob bins

dwelch-spike:


dwelch-spike:
working on test fix...

jdogmcsteezy:
I would follow the server convention in terms of branch names. That is what I did except I used bugfix instead of hotfix since that is how we refer to it. You should use `bugfix/4.1`. It is nice in some GUIs because it will actually organize it into a folder structure

image

.github/workflows/backup-build.yml
```diff
@@ -2,9 +2,9 @@ name: Build and Test All
 
 on:
   push:
-    branches: [ master, test-ready ]
+    branches: [ master, test-ready, "bugfix-*" ]
```
jdogmcsteezy: The `bugfix-*` looks like a duplicate of what I use in asadm and maybe uda/aql. If it is, I don't normally use them in this way. Normally I am not merging into bugfix but merging bugfix into master. I did not do that this time though

.github/workflows/mac-artifact.yml
```diff
@@ -1,9 +1,9 @@
 name: Mac Artifact
 on:
   push:
-    branches: [ master, test-ready, "bugfix-*", actionsHub ]
+    branches: [ master, test-ready, "bugfix-*" ]
   pull_request:
-    branches: [ master, test-ready ]
+    branches: [ master, test-ready, "bugfix-*" ]
```
jdogmcsteezy, line 6: I would just check that the builds are not happening twice.

README.md
```diff
@@ -317,7 +317,7 @@ Let's look at the placeholders, there are quite a few.
 | `{name}`       | The name of the index. |
 | `{index-type}` | The type of index: `N` = index on bins, `L` = index on list elements, `K` = index on map keys, `V` = index on map values |
 | `{path}`       | The bin name |
-| `{data-type}`  | The data type of the indexed value: `N` = numeric, `S` = string |
+| `{data-type}`  | The data type of the indexed value: `N` = numeric, `S` = string, `G` = geo2dsphere, `B` = bytes/blob, `I` = invalid |
```
jdogmcsteezy, line 320: When would it be invalid?

jdogmcsteezy (approved):
Couple comments but LGTM

README.md
dwelch-spike: This shouldn't ever happen, but it's in the code as a default case.

dwelch-spike:
> I would follow the server convention in terms of branch names. That is what I did except I used bugfix instead of hotfix since that is how we refer to it. You should use `bugfix/4.1`. It is nice in some GUIs because it will actually organize it into a folder structure
> image

I got this from https://aerospike.atlassian.net/wiki/spaces/TOOLS/pages/2793472010/Tools+Development+Git+Workflow where we say to use "bugfix-x.y.z" where x.y.z is the version the bugfix will be released in. I'm fine changing this though, I'll change it in this doc to be "bugfix/x.y.z" where x.y.z is the version being patched.

.github/workflows/mac-artifact.yml
dwelch-spike: I don't see duplicate builds in the actions run logs

.github/workflows/backup-build.yml
dwelch-spike: Yeah I'll remove this so only PRs will trigger the runs

PR #76, opened 2023-11-28
TOOLS-2779 Do not try to set S3 configuration if initialized

reugn:


dwelch-spike:
@reugn Thanks for the PR. Looks like the tests are failing due to some kind of connection issue. I'll take a look at them soon. I want to fix them before merging this.

dwelch-spike:
LGTM

PR #78, opened 2023-12-05
tools-2788 Reset s3API state on shutdown.

dwelch-spike:
This clears out state in between s3 backups.

### DIFF
--- a/src/s3_api.cc
+++ b/src/s3_api.cc
@@ -63,6 +63,9 @@ S3API::TryInitialize()
 void
 S3API::Shutdown()
 {
+	// TODO this lock probably doesn't make sense
+	// if this object is being shutdown, then no other threads
+	// should be using it
 	std::unique_lock<std::mutex> lg(s3_init_lock);
 	if (initialized) {
 		inf("Closing S3 API");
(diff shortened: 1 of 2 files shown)


src/s3_api.cc
```diff
@@ -71,6 +74,18 @@ S3API::Shutdown()
 		}
 
 		Aws::ShutdownAPI(options);
+
+		this->region.clear();
+		this->bucket.clear();
+		this->profile.clear();
+		this->endpoint.clear();
+		// NOTE: logLevel should be reset to match
+		// whatever S3_DEFAULT_LOG_LEVEL is if it changes
```
jdogmcsteezy, line 83: Is this a TODO or a note? It does not look like you are doing what it is recommending

jdogmcsteezy (commented):
Just a Q

src/s3_api.cc
dwelch-spike: It's a reminder to update this if the default changes in the future. I can't use the default that asbackup configs use because it is a different type

jdogmcsteezy (approved):

PR #79, opened 2023-12-06
fix: tools-2751 use effective replication factor when estimating back…

dwelch-spike:
…up size

### DIFF
--- a/src/backup_status.c
+++ b/src/backup_status.c
@@ -1508,7 +1508,10 @@ get_object_count(aerospike *as, const char *namespace, as_vector* set_list,
 	}
 
 	// if effective replication factor is 0 set it to 1 to prevent division by 0
-	effective_repl_factor = effective_repl_factor == 0 ? 1 : effective_repl_factor;
+	if (effective_repl_factor == 0) {
+		inf("Warning: effective replication factor is 0");
+		effective_repl_factor = 1;
+	}
 
 	*obj_count /= effective_repl_factor;
 
--- a/src/backup_status.c
+++ b/src/backup_status.c
@@ -1326,7 +1326,7 @@ static bool
 ns_count_callback(void *context_, const char *key, const char *value)
 {
 	ns_count_context *context = (ns_count_context *)context_;
-	int64_t repl_factor;
+	int64_t effective_repl_factor;
 	int64_t object_count;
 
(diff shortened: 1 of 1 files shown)


src/backup_status.c
```diff
@@ -1339,13 +1339,13 @@ ns_count_callback(void *context_, const char *key, const char *value)
 		return true;
 	}
 
-	if (strcmp(key, "replication-factor") == 0) {
-		if (!better_atoi(value, &repl_factor) || repl_factor < 0 || repl_factor > 100) {
+	if (strcmp(key, "effective_replication_factor") == 0) {
+		if (!better_atoi(value, &effective_repl_factor) || effective_repl_factor < 0 || effective_repl_factor > 100) {
```
jdogmcsteezy: effective_repl_factor can be 0 in a fresh SC cluster without the roster set. I would check for that case as well.

jdogmcsteezy (changes requested):

src/backup_status.c
dwelch-spike: Note to self... a effective_repl_factor of 0 will lead to a divide by 0 error. The max effective_repl_factor should be 256.

dwelch-spike: Changed this so that if EFR is set to 0, `obj_count` is divided by 1 to avoid a divide by 0.

Set the max check for EFR from 100 to 256.

dwelch-spike:
@jdogmcsteezy please have another look when you can.

src/backup_status.c
```diff
@@ -1507,6 +1507,9 @@ get_object_count(aerospike *as, const char *namespace, as_vector* set_list,
 		*obj_count += count;
 	}
 
+	// if effective replication factor is 0 set it to 1 to prevent division by 0
+	effective_repl_factor = effective_repl_factor == 0 ? 1 : effective_repl_factor;
```
jdogmcsteezy: What's the goal of changing it to 1?  I would think a warning like the one you have above would be a better approach. Odds are the estimate is going to be wrong if object count != 0 since the effective replication factor was likely > 1 at one point

dwelch-spike: It's to prevent a divide by 0 error. This value is divided into obj_count later.

A warning is a good idea. I'll add one.

jdogmcsteezy (approved):

PR #81, opened 2023-12-07
TOOLS-2792 Add default AWS credentials provider

reugn:


reugn:
The default provider is already in use.

PR #83, opened 2023-12-16
tools-2802 fix: only use async clients to prevent teardown memory leak in shared…

dwelch-spike:
… lib


jdogmcsteezy (approved):
LGTM, the leak was caused by subsequent calls without teardown where one client is sync and the other is async. On a subsequent call the client destructor thinks both clients are async causing heap memory to be left unfreed

PR #84, opened 2023-12-18
Tools 2804

dwelch-spike:
Fixes small config memory leaks when asbackup/restore is used as a shared lib. These occur when users overwrite heap allocated config default values.


dwelch-spike:
Needed by the backup service team, merging.

PR #86, opened 2024-01-16
feat: tools-2646 add --prefer-racks flag

dwelch-spike:


src/backup_status.c
```diff
@@ -348,6 +348,32 @@ backup_status_init(backup_status_t* status, backup_config_t* conf)
...
+				err("Invalid rack id %s", id_str);
+
+				cf_free(racks_clone);
+				as_vector_destroy(&rackids);
+
+				goto cleanup2;
+			}
+
+			as_config_add_rack_id(&as_conf, (int)id);
+		}
```
jdogmcsteezy, line 371: Do we have a way to test this?

dwelch-spike: I'm not sure how to tell that it is active from a test. I just checked that the value gets set in the C client policy from a debugger.

dwelch-spike: I can still exercise this part of the code in an integration test though. I'll add it to one.

jdogmcsteezy (approved):

PR #92, opened 2024-04-24
fix: tools-2902 count ignored records caused by exceeding max retries

dwelch-spike:


dwelch-spike:
This might change so I'm making it a draft for now.

PR #101, opened 2024-07-12
fix: tools-2941 properly upload state files to s3.

dwelch-spike:
State files could not be uploaded to S3 due to a race condition where the upload manager was told to stop uploads before the state file could be written to s3.


dwelch-spike:
Mac build is failing because macos-11 runners are no longer supported. I'll deal with that in another PR.

jdogmcsteezy (approved):
LGTM. Looks like it adds a check to make sure the user did not do something they weren't supposed to

PR #102, opened 2024-07-30
build: add runner.arch cache key

dwelch-spike:


dwelch-spike:
This should fix the cmake cache issues causing mac builds to fail.

PR #110, opened 2025-06-10
fix: TOOLS-3050 fix stack corruption happening due to buffer overflow.

akash-aero:
This fix resolves a buffer overrun vulnerability that could lead to memory corruption.
str_vector_tostring() could overflow the buffer when too many strings were appended without bounds checking.

**Fix**:  
we've Introduced length checks before each append and added truncation with "..." when the buffer limit is reached.

**Expected in logs**:
If the list is too long, you'll now see a truncated output ending with ..., and no buffer overflow crashes or corrupted output.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -357,15 +357,29 @@ str_vector_tostring(const as_vector* v)
 
 	if (v->size == 0) {
 		buf[0] = '\0';
+		return buf;
 	}
 
+	char* dots = "...";
+	uint32_t needed = 0;
+
 	for (uint32_t i = 0; i < v->size; i++) {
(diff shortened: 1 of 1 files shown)


src/utils.c
```diff
@@ -2282,4 +2296,4 @@ read_private_key(char* pkey_data,
 		return -1;
 	}
 	return 0;
-}
```
srivarun91, line 2285: why is this showing up?

akash-aero: Showing  extra space, I've removed it now.

akash-aero:
[asbackup_test_logs.txt](https://github.com/user-attachments/files/20722747/asbackup_test_logs.txt)

srivarun91 (approved):
Pair coded.

PR #113, opened 2025-09-17
fix: TOOLS-3127 reclaim s3 download slots when downloads are finished instead of consumed ops-7812...

dwelch-spike:
...preventing deadlock when consumer is waiting on a part that is not downloading when max downloads are in progress ops-7812

### DIFF
--- a/src/download_manager.cc
+++ b/src/download_manager.cc
@@ -182,8 +182,6 @@ DownloadManager::AwaitAllDownloads()
 	outstanding_calls_cv.wait(lg, [&]() { return this->outstanding_calls == 0; });
 	lg.unlock();
 
-	// Mark all of the parts in result_queue as complete
-	gdm->PartsComplete(false, uint32_t(result_queue.get_n_items()));
 	return true;
 }
 
@@ -219,7 +217,9 @@ void
 DownloadManager::DecrementOutstandingCalls()
 {
 	outstanding_calls_lock.lock();
-	outstanding_calls--;
+	if (outstanding_calls > 0) {
+		outstanding_calls--;
+	}
 	outstanding_calls_lock.unlock();
 	outstanding_calls_cv.notify_one();
 }
@@ -243,11 +243,16 @@ DownloadManager::GetObjectFinished(
 	uint64_t download_idx = async_ctx->GetPartN();
 
 	if (!outcome.IsSuccess()) {
-		err("%s", outcome.GetError().GetMessage().c_str());
-		received_cb(std::const_pointer_cast<Aws::Client::AsyncCallerContext>(context), false);
+    	err("DM download error part=%llu: %s",
+			(unsigned long long)download_idx,
+			outcome.GetError().GetMessage().c_str());
 
 		download_manager->result_queue.push_error(download_idx);
 		download_manager->SetError();
+
+		// Decrement the group download manager's outstanding calls
+		received_cb(std::const_pointer_cast<Aws::Client::AsyncCallerContext>(context), false);
+		// Decrement this manager's ongoing download count
 		download_manager->DecrementOutstandingCalls();
 		return;
 	}
@@ -266,8 +271,8 @@ bool
 DownloadManager::AwaitDownloadPart()
 {
 	std::pair<Aws::S3::Model::GetObjectResult, uint64_t>* res = result_queue.pop();
-	g_api.GetGroupDownloadManager()->PartComplete(true);
-	if (res == nullptr) {
+
+    if (res == nullptr) {
 		SetError();
 		return false;
 	}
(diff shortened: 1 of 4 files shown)


src/download_manager.cc
```diff
@@ -320,39 +325,16 @@ GroupDownloadManager::RemoveDownloadManager(DownloadManager* dm)
...
 void
 GroupDownloadManager::PartDownloadComplete(GroupDownloadManager* gdm,
 		std::shared_ptr<Aws::Client::AsyncCallerContext> context,
 		bool success)
 {
 	(void) context;
 
+	std::unique_lock<std::mutex> lg(gdm->access_lock);
+	gdm->async_downloads--;
 	if (success) {
```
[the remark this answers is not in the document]
dwelch-spike: The old PartComplete methods also decremented async_downloads in failure cases, this is intended.

src/download_manager.cc
```diff
@@ -219,7 +217,9 @@ void
 DownloadManager::DecrementOutstandingCalls()
 {
 	outstanding_calls_lock.lock();
-	outstanding_calls--;
+	if (outstanding_calls > 0) {
+		outstanding_calls--;
+	}
```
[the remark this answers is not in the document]
dwelch-spike: outstanding_calls_lock.lock(); takes the lock before the if

src/download_manager.cc
```diff
@@ -243,11 +243,16 @@ DownloadManager::GetObjectFinished(
 	uint64_t download_idx = async_ctx->GetPartN();
 
 	if (!outcome.IsSuccess()) {
-		err("%s", outcome.GetError().GetMessage().c_str());
-		received_cb(std::const_pointer_cast<Aws::Client::AsyncCallerContext>(context), false);
+    	err("DM download error part=%llu: %s",
```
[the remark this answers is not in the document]
dwelch-spike: fixed

reugn (approved):

src/download_manager.cc
```diff
@@ -320,39 +325,16 @@ GroupDownloadManager::RemoveDownloadManager(DownloadManager* dm)
...
 {
 	(void) context;
 
+	std::unique_lock<std::mutex> lg(gdm->access_lock);
+	gdm->async_downloads--;
 	if (success) {
-		std::unique_lock<std::mutex> lg(gdm->access_lock);
 
 		// for exponential rampup, try initiating two downloads
 		gdm->StartNextPart();
```
a-spiker, line 340: nitpick: should we log if StartNextPart() returns false ?

src/download_manager.cc
```diff
@@ -320,39 +325,16 @@ GroupDownloadManager::RemoveDownloadManager(DownloadManager* dm)
...
-		}
-	}
-}
-
 void
 GroupDownloadManager::PartDownloadComplete(GroupDownloadManager* gdm,
 		std::shared_ptr<Aws::Client::AsyncCallerContext> context,
 		bool success)
 {
 	(void) context;
```
a-spiker, line 333: ntipick:
```suggestion
	std::shared_ptr<const StreamManager::AsyncContext> async_ctx =
		std::static_pointer_cast<const StreamManager::AsyncContext>(context);
	uint64_t download_idx = async_ctx->GetPartN();
	
	std::unique_lock<std::mutex> lg(gdm->access_lock);
	gdm->async_downloads--;
	
	ver("S3 download progress: %u/%u active downloads, part %llu completed %s", 
		gdm->async_downloads, gdm->max_async_downloads, 
		(unsigned long long)download_idx, success ? "successfully" : "with error");
```

How about logging here ? 
```
[VER] S3 download progress: 15/32 active downloads, part 42 completed successfully
[VER] S3 download progress: 14/32 active downloads, part 43 completed with error
```

a-spiker (commented):
Minor nitpicks, I'm good with the changes

src/download_manager.cc
dwelch-spike: I think something like this isn't a bad idea but I have a few concerns/thoughts... I think debug/verbose logging improvements for asrestore/asbackup should be a separate work item.
1. ver() verbose logging does not work in the c++ files, this is because the global variables used to implement verbose logging `g_verbose` are c11 atomics that c++ doesn't understand. We can work around this by wrapping them but I think that's beyond the scope of this PR.
2. logging here would produce quite a bit of output and we should probably try to associate the log with the object being downloaded. We should think about what we get out of this log and if there is a way to get the same information with less frequent log output, for example, I think checking the file open log lines for associated file closed log lines can tell us when hangs like this exist already, for example, if asrestore hangs without closing certain files, we know that the downloads for those files may be deadlocked.

src/download_manager.cc
dwelch-spike: I think it only returns false when an object is already finished downloading. I don't think it's worth logging over. I do think adding some more detailed debug logging in the s3 logic is a good idea, but I'm not sure what, and we need to make verbose logging work first.

PR #118, opened 2026-01-22
chore: update git describe command to use --abbrev=9 for versioning in Makefile and workflow

a-spiker:
This pull request updates how git version tags are generated in both the GitHub Actions workflow and the Makefile. The main change is to use the `--abbrev=9` option with `git describe`, which ensures that abbreviated commit hashes are consistently 9 characters long. This helps standardize version strings across builds.

Versioning improvements:

* Updated the `git describe` command in `.github/workflows/mac-artifact.yml` to use `--abbrev=9`, ensuring that generated tags include a 9-character abbreviated commit hash.
* Updated the `VERSION` variable in the `Makefile` to use `git describe --abbrev=9`, standardizing version output for local builds as well.

### DIFF
--- a/.github/workflows/mac-artifact.yml
+++ b/.github/workflows/mac-artifact.yml
@@ -211,8 +211,8 @@ jobs:
       working-directory: ${{ steps.working-dir.outputs.value }}
       id: tag
       run: |
-        git describe --tags --always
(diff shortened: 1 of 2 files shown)


pvinh-spike (approved):
LGTM

pvinh-spike (approved):
LGTM

PR #119, opened 2026-02-04
fix: TOOLS-3183 Add macOS x64 build support

a-spiker:
Add macos-15-intel runner to the build matrix with x86_64-specific paths for all dependencies (openssl, zstd, aws-sdk, curl, ssh2, uv, jansson) at /usr/local prefix.


pvinh-spike (approved):
LGTM

PR #135, opened 2026-05-04
fix: TOOLS-3510 Backup file extension truncated from .asb to .as when count over 100k

volmarl:
When count of backup files exceeds 100000 the backup file name is truncated from .asb to .as This affects asrestore lookign for .abs

By using file_count and the %05" PRId64 " specifier in the sizing call:

Correct Calculation: When file_count is 100,000, snprintf will accurately report that it needs space for 6 digits instead of 5.  

Sufficient Allocation: The subsequent cf_malloc((file_path_size + 1)) will then allocate enough memory for the full .asb extension.  

No Truncation: The final snprintf (around line 1400) will no longer drop the 'b' to fit the string into the buffer.

### DIFF
--- a/src/backup.c
+++ b/src/backup.c
@@ -1395,6 +1395,7 @@ open_dir_file(backup_job_context_t *bjc)
 			pthread_mutex_unlock(&bjc->status->dir_file_init_mutex);
 			cf_free(bjc->fd);
 			bjc->fd = NULL;
+			err("Unable to malloc file path name");
 			return false;
 		}
 
--- a/src/backup.c
+++ b/src/backup.c
@@ -1394,13 +1394,12 @@ open_dir_file(backup_job_context_t *bjc)
 
 		char* file_path = (char*) cf_malloc((file_path_size + 1) * sizeof(char));
 		if (file_path == NULL) {
+			pthread_mutex_unlock(&bjc->status->dir_file_init_mutex);
 			cf_free(bjc->fd);
 			bjc->fd = NULL;
 			err("Unable to malloc file path name of length %" PRIu64, file_path_size);
 			return false;
 		}
-
-
 
 		snprintf(file_path, file_path_size + 1, "%s/%s_%05" PRId64 ".asb",
 				bjc->conf->directory,
--- a/src/backup.c
+++ b/src/backup.c
@@ -1384,6 +1384,9 @@ open_dir_file(backup_job_context_t *bjc)
 			return false;
 		}
 
+		pthread_mutex_lock(&bjc->status->dir_file_init_mutex);
+		int64_t file_count = bjc->status->file_count;
+
 		uint64_t file_path_size = (size_t) snprintf(NULL, 0, "%s/%s_%05" PRId64 ".asb",
 				bjc->conf->directory,
 				bjc->conf->prefix == NULL ? bjc->conf->ns : bjc->conf->prefix,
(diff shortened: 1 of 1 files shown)


src/backup.c
```diff
@@ -1397,8 +1400,7 @@ open_dir_file(backup_job_context_t *bjc)
 			return false;
```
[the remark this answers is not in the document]
dwelch-spike: As copilot says this can lead to deadlock, need to release the lock in error cases.

src/backup.c
```diff
@@ -1397,8 +1400,7 @@ open_dir_file(backup_job_context_t *bjc)
...
 		}
 
-		pthread_mutex_lock(&bjc->status->dir_file_init_mutex);
-		int64_t file_count = bjc->status->file_count;
+
```
dwelch-spike: nit: don't need this extra empty line.

dwelch-spike (changes requested):
Need to release the lock before returning if malloc fails and could remove that extra blank line. After that this will LGTM. Let's make sure this gets a little testing too.

src/backup.c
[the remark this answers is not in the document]
volmarl: PR updated with mutex unlocked

dwelch-spike (dismissed):
This LGTM.

dwelch-spike:
@a-spiker please make sure QE knows we merged a change that could affect backup file names, aslong as they have tests that produce large numbers of backup files that should be enough regression testing.

src/backup.c
```diff
@@ -1384,22 +1384,23 @@ open_dir_file(backup_job_context_t *bjc)
...
+		pthread_mutex_lock(&bjc->status->dir_file_init_mutex);
+		int64_t file_count = bjc->status->file_count;
+
+		uint64_t file_path_size = (size_t) snprintf(NULL, 0, "%s/%s_%05" PRId64 ".asb",
 				bjc->conf->directory,
 				bjc->conf->prefix == NULL ? bjc->conf->ns : bjc->conf->prefix,
-				0);
+				file_count);
 
 		char* file_path = (char*) cf_malloc((file_path_size + 1) * sizeof(char));
```
a-spiker:
```suggestion
		char* file_path = dyn_sprintf("%s/%s_%05" PRId64 ".asb",
        bjc->conf->directory,
        bjc->conf->prefix == NULL ? bjc->conf->ns : bjc->conf->prefix,
        file_count);
```

 I recommend doing this to ensure the bug being fixed cannot recur because there's no second format string to fall out of sync.

a-spiker (commented):
Minor comments

src/backup.c
```diff
@@ -1430,8 +1418,10 @@ open_dir_file(backup_job_context_t *bjc)
 			cf_free(bjc->fd);
 			bjc->fd = NULL;
 			err("New directory file %s, failed to get file position", file_path);
+			cf_free(file_path);
```
dwelch-spike, line 1422: Good catch, this fixes a use after free of file_path.

src/backup.c
```diff
@@ -1387,28 +1387,17 @@ open_dir_file(backup_job_context_t *bjc)
...
-
-		char* file_path = dyn_sprintf("%s/%s_%05" PRId64 ".asb",
-        bjc->conf->directory,
-        bjc->conf->prefix == NULL ? bjc->conf->ns : bjc->conf->prefix,
-        file_count);
 		if (file_path == NULL) {
 			pthread_mutex_unlock(&bjc->status->dir_file_init_mutex);
 			cf_free(bjc->fd);
 			bjc->fd = NULL;
-			err("Unable to malloc file path name of length %" PRIu64, file_path_size);
```
dwelch-spike, line 1396: I think we should still have an error log here, something like "Unable to construct backup file name".

dwelch-spike (changes requested):
This is a good improvement and fixes a potential use after free. I think we should still have an error log if we fail to create the backup file name though.

dwelch-spike (approved):
LGTM

PR #136, opened 2026-05-15
feat: TOOLS-3545 add --s3-allow-system-proxy to honor HTTP(S)_PROXY env vars

a-spiker:
## Summary

Implements [TOOLS-3545](https://aerospike.atlassian.net/browse/TOOLS-3545). Adds `--s3-allow-system-proxy` to `asbackup` and `asrestore` so the S3 client honors the standard system proxy environment variables (`HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` / `NO_PROXY` and lowercase variants). Off by default — behavior is unchanged for users not behind a proxy.

Customer (Experian) is on a host whose only path to S3 is a corporate HTTP proxy advertised via `HTTPS_PROXY`. With the flag set, libcurl picks up the env var and the backup completes.

## Approach

The PR uses `ClientConfiguration::allowSystemProxy = true`, which the AWS SDK forwards to libcurl. libcurl then handles env-var precedence, `NO_PROXY` matching (including wildcards), Basic-auth credentials, and SOCKS schemes natively.

`allowSystemProxy` was added in `aws-sdk-cpp` 1.11.169, so this PR also bumps the pinned SDK from **1.10.55 → 1.11.813** everywhere it's referenced.

An earlier iteration of this branch shipped a homegrown URL parser + env-var pickers + a separate `s3_proxy` module. That whole approach was dropped in favor of the SDK-native flag — see the `refactor:` commit. Net result is ~1000 lines deleted vs. the intermediate state.

### Feature plumbing (`--s3-allow-system-proxy`)

CLI longopt + TOML key + config struct field + clone/init/help, identical pattern to other `--s3-*` flags:

- [`src/backup_config.c`](src/backup_config.c), [`src/restore_config.c`](src/restore_config.c) — longopt, switch case, init default (`false`), clone (backup), help text
- [`src/conf.c`](src/conf.c) — TOML `s3-allow-system-proxy = true/false` for `[asbackup]` and `[asrestore]`
- [`include/backup_config.h`](include/backup_config.h), [`include/restore_config.h`](include/restore_config.h), [`include/utils.h`](include/utils.h) — field + enum
- [`src/backup.c`](src/backup.c), [`src/restore.c`](src/restore.c) — `set_s3_configs` propagates the flag
- [`include/file_proxy.h`](include/file_proxy.h), [`src/file_proxy_s3.cc`](src/file_proxy_s3.cc) — `s3_set_allow_system_proxy()` wrapper
- [`include/s3_api.h`](include/s3_api.h), [`src/s3_api.cc`](src/s3_api.cc) — `S3API::SetAllowSystemProxy`, `allow_system_proxy` member (reset in `Shutdown`), assigned to `conf.allowSystemProxy` in `_init_api`

### Pre-existing bug fix bundled in

[`src/s3_api.cc:309-317`](src/s3_api.cc#L309-L317) — `conf.scheme = HTTP` was hardcoded whenever `--s3-endpoint-override` was set, regardless of whether the override URL was `http://` or `https://`. The SDK uses the scheme prefix from `endpointOverride` directly so actual traffic was correct, but `conf.scheme` was inconsistent. Now derived from the URL prefix (case-insensitive). Bare `host:port` still defaults to HTTP — existing tests unaffected.

### SDK 1.11 compatibility

- [`include/asbackup_logger.h`](include/asbackup_logger.h), [`src/asbackup_logger.cc`](src/asbackup_logger.cc) — added the `vaLog` override required by 1.11's `FormattedLogSystem` abstract base; `Log(...)` now `va_start`s and delegates
- [`Makefile`](Makefile) — link `-framework Network` on Darwin (`aws-c-io` in 1.11 uses Apple's Network.framework on macOS)

### Build / installer pipeline

- [`.github/workflows/backup-build.yml`](.github/workflows/backup-build.yml), [`.github/workflows/mac-artifact.yml`](.github/workflows/mac-artifact.yml) — `AWS_SDK_CPP_VERSION: 1.11.813`
- [`.github/packaging/project/install_deps.sh`](.github/packaging/project/install_deps.sh) — single source of truth (`AWS_SDK_VERSION="1.11.813"`) drives both `.deb` and `.rpm` builds. Also converted all 10 `git clone` sites to `--depth 1 --branch $AWS_SDK_VERSION` + shallow submodule init (aws-sdk-cpp pulls in ~10 CRT submodules; shallow is a substantial speedup).
- [`Makefile`](Makefile) — minor: `VERSION ?=` so CI can inject SemVer before a tag exists; test rule now declares `$(C_CLIENT_LIB) $(SECRET_CLIENT_LIB) $(TOML)` as prereqs explicitly

### Hygiene

- [`.gitignore`](.gitignore) — `*.asb.state`, `/bin` anchor, docker bake artifacts

### Unit ([`test/unit/test_backup_conf.c`](test/unit/test_backup_conf.c), [`test/unit/test_restore_conf.c`](test/unit/test_restore_conf.c))

Round-trip config tests for the new flag — TOML parse, init, clone equality.

### Integration ([`test/integration/test_s3.py`](test/integration/test_s3.py))

Three new tests using a bind-then-close dead port as the unreachable proxy:

| Test | Verifies |
|---|---|
| `test_s3_allow_system_proxy` | Without flag + dead `HTTP_PROXY`/`HTTPS_PROXY` → backup reaches MinIO direct → succeeds. With flag → routes through dead proxy → fails. Proves the flag is what flips the SDK into honoring the env. |
| `test_s3_allow_system_proxy_restore` | Same two-part assertion on restore. AC explicitly requires both backup AND restore to succeed via `HTTPS_PROXY`. |
| `test_s3_allow_system_proxy_no_proxy_exempts_host` | With flag + dead proxy + `NO_PROXY=127.0.0.1` → backup succeeds. End-to-end proof that `NO_PROXY` is honored. Mirrors realistic customer scenario (external via proxy, internal direct). |

Helper `_dead_proxy_env()` sets all six lowercase/uppercase `HTTP/HTTPS/ALL_PROXY` env vars to a random closed local port and clears `NO_PROXY`; accepts `extra=` for per-test overrides.

## User-facing usage

```bash
export HTTPS_PROXY=http://corp-proxy.internal:3128
export NO_PROXY=.internal.corp,127.0.0.1,169.254.169.254

asbackup -n test -d s3://bucket/path --s3-allow-system-proxy
asrestore -n test -d s3://bucket/path --s3-allow-system-proxy
```

TOML:
```toml
[asbackup]
s3-allow-system-proxy = true

[asrestore]
s3-allow-system-proxy = true
```

Supported env-var precedence, URL syntax (including IPv6, userinfo, SOCKS), `NO_PROXY` matching, and httpoxy handling all follow libcurl behavior — see [libcurl docs](https://curl.se/libcurl/c/CURLOPT_PROXY.html).

## Ticket acceptance criteria

| AC | Status |
|---|---|
| New flag + config key documented & validated like other `--s3-*` options | ✅ longopt, TOML, init/clone, help text, unit round-trip |
| `ClientConfiguration::allowSystemProxy = true` applied before client init | ✅ direct SDK field |
| E2E verified: backup AND restore succeed via `HTTPS_PROXY` | ✅ both tested |

## Test plan

- [x] `make unit` — config round-trip passes
- [x] `make run_test_s3` — 3 new proxy tests pass against MinIO
- [x] CI build + .deb + .rpm + mac artifacts (validates SDK 1.11.813 bump everywhere)
- [ ] Manual: customer-style scenario (host behind HTTP proxy advertised via `HTTPS_PROXY`, real S3)

[TOOLS-3545]: https://aerospike.atlassian.net/browse/TOOLS-3545


dwelch-spike:
This is a lot of new code, @a-spiker did you look at upgrading the aws-sdk? The description mentions "The ticket asks for ClientConfiguration::allowSystemProxy = true. That field does not exist in aws-sdk-cpp 1.10.55 (the version CI pins) — it landed with the CRT-integrated 1.11.x line." Maybe upgrading to 1.11 would enable a much smaller code change? Would need to look at the aws sdk release notes to make sure an upgrade is safe.

dwelch-spike:
Took a careful look at this with the SDK side in mind. The diagnosis is correct — `ClientConfiguration::allowSystemProxy` does not exist in 1.10.55. I traced it to commit [`a3a39d04f5`](https://github.com/aws/aws-sdk-cpp/commit/a3a39d04f501726652acd4524eb5988c315500a0) ("Allow usage of system default proxy"), which first appears in tag **1.11.169** (Sep 2023). The implementation matches what's described inline: when the flag is true, the SDK skips `curl_easy_setopt(handle, CURLOPT_PROXY, "")` and libcurl reads the env vars natively.

That means a bump to ≥ 1.11.169 would replace `src/s3_proxy.{h,cc}` (~600 LOC of parsing + ~250 LOC of tests) with a single `conf.allowSystemProxy = ...` line. Worth weighing against the current approach.

**Source-compatibility audit against 1.11.169** (tag-diffed, for each SDK surface this repo touches):

| Surface | Verdict |
|---|---|
| `S3Client(ClientConfig, PayloadSigningPolicy, bool)` ctor at `src/s3_api.cc:309` | unchanged |
| Sync ops (`ListObjectsV2`, `ListMultipartUploads`, `HeadObject`, `DeleteObjects`, `AbortMultipartUpload`) | unchanged |
| `client.UploadPartAsync` / `GetObjectAsync` (1.11.0 made these non-virtual templates) | upstream's own guidance: no source change needed unless you subclass `S3Client` and override async methods — we don't |
| `Aws::Client::AsyncCallerContext` | byte-identical |
| `Aws::Utils::Logging::LogSystemInterface` (`AsbackupLogger`'s actual parent) | unchanged |
| `vaLog` on `FormattedLogSystem` | added *between 1.11.169 and 1.11.500*; `AsbackupLogger` doesn't inherit from `FormattedLogSystem` so this doesn't bind us either way |
| `Aws::SDKOptions` | additive (new `sdkVersion` substruct) |

So the upgrade is source-compatible — no code changes other than deleting `s3_proxy.*`.

**Costs the SDK bump would have that the current PR avoids:**

1. CI churn — `.github/workflows/{backup-build,mac-artifact}.yml` pin `AWS_SDK_CPP_VERSION: 1.10.55`. Bumping invalidates the cached-SDK build matrix and re-runs everything mac/linux.
2. ~620 patch releases of drift since 1.11.169 (retry defaults, IMDSv2, endpoint resolution, request compression). Even bug-for-bug compatible at the API surface, runtime behavior of every S3 op shifts more than a focused proxy patch does.
3. The whole `Makefile` / `README.md` / `vaLog` defense in this PR is a 1.10.55 pin — it disappears with an upgrade.

**Two ways forward I see — what do you think?**

1. **Slim this PR**: keep the diagnosis and the pre-existing scheme-inference bug fix at `src/s3_api.cc:299-302` (good catch — that's a real bug independent of proxy work), but replace the hand-rolled URL parser with `Aws::Http::URI` (already in 1.10.55). Httpoxy / NO_PROXY tokenization / IPv6 brackets logic shrinks to env-var picking + URI field reads. Keeps the 1.10.55 pin, ships fast for Experian.
2. **Bump SDK to ≥ 1.11.169**: deletes `s3_proxy.*` entirely. Source-compatible per above. Costs a cross-platform CI re-validation and accepts ~2 years of SDK drift in one go — but the long-term answer is fewer lines we own.

If Experian has a hard deadline, (1) ships faster. If we have CI room, I'd lean toward (2) — future proxy fixes (CIDR NO_PROXY, port-specific entries — both already listed under known limitations here) come for free from upstream as they fix them.

Either way, nice writeup in the PR description — the inline citation of `CurlHttpClient.cpp:718` made this much easier to verify.

src/s3_proxy.cc
```diff
@@ -0,0 +1,267 @@
...
+		std::string decoded_user;
+		std::string decoded_pass;
+		if (!percent_decode(raw_user, decoded_user) ||
+				!percent_decode(raw_pass, decoded_pass)) {
+			std::memset(out, 0, sizeof(*out));
+			return S3_PROXY_PARSE_MALFORMED;
+		}
+		if (!copy_to_buf(decoded_user, out->username, sizeof(out->username)) ||
+				!copy_to_buf(decoded_pass, out->password,
+						sizeof(out->password))) {
```
dwelch-spike: From Data flow agent: **Important:** `percent_decode` accepts `%00`, which puts a literal NUL inside `decoded_user` / `decoded_pass`. `copy_to_buf` then `memcpy`s the embedded NUL into the fixed buffer, and the sink at `src/s3_api.cc:356-361` checks `parsed.username[0] != '\0'` and assigns `Aws::String` from `const char*` (i.e. via `strlen`). Net effect: a credential whose first decoded byte is NUL is silently dropped, and one with `%00` later in the string is silently truncated — proxy auth then sends with no credential or a short credential rather than failing loudly. Reject `%00` at parse time inside `percent_decode` so the user gets a clear `malformed proxy env var` error instead of a silent auth misconfiguration.

src/s3_proxy.cc
```diff
@@ -0,0 +1,267 @@
...
+		return S3_PROXY_PARSE_MALFORMED;
+	}
+	if (!copy_to_buf(host, out->host, sizeof(out->host))) {
+		std::memset(out, 0, sizeof(*out));
+		return S3_PROXY_PARSE_MALFORMED;
+	}
+
+	unsigned long port_val;
+	if (port_str.empty()) {
+		port_val = is_https ? 443UL : 80UL;
```
dwelch-spike: From Data flow agent: **Minor:** Inputs like `http://proxy:` or `http://[::1]:` (trailing colon, no digits) take the `port_str.empty()` branch and silently use the scheme default port (80/443). That's not what the user typed, and libcurl would reject the bare trailing colon anyway. Track whether a colon was actually present in the authority split and require non-empty digits when it was — return `S3_PROXY_PARSE_MALFORMED` so misconfigured env vars don't connect to an unintended port.

test/unit/test_s3_proxy.c
```diff
@@ -0,0 +1,487 @@
...
+{
+	s3_proxy_url_t r;
+	ck_assert_int_eq(s3_proxy_parse_url("http://[::1", &r),
+			S3_PROXY_PARSE_MALFORMED);
+	ck_assert_int_eq(s3_proxy_parse_url("http://[]", &r),
+			S3_PROXY_PARSE_MALFORMED);
+}
+END_TEST
+
+START_TEST(parse_bad_percent_encoding_malformed)
```
dwelch-spike: From Tests agent: **Minor:** Both assertions in `parse_bad_percent_encoding_malformed` put the bad escape inside the password. The implementation calls `percent_decode` separately on `raw_user` and `raw_pass` (and ORs the results), so a regression that broke only the username decode path would pass this test. Add a case like `http://%ZZ:x@h` (which the PR description explicitly lists as documented behavior) to cover username-side bad escapes.

src/s3_api.cc
```diff
@@ -298,13 +312,82 @@ S3API::_init_api(S3API& s3_api)
...
+			}
+		}
+
+		const char* no_proxy = s3_proxy_pick_no_proxy_env();
+		if (no_proxy != nullptr) {
+			Aws::Vector<Aws::String> hosts;
+			std::string np(no_proxy);
+			size_t i = 0;
+			while (i < np.size()) {
+				size_t j = np.find_first_of(", \t\r\n", i);
```
dwelch-spike: From Tests agent: **Minor:** The NO_PROXY tokenizer at lines 374-384 is the only place a typo or off-by-one in the host-list parse would surface, and no unit test exercises it. The integration tests only ever pass a single host. A unit-level test asserting that `"a.com, b.com\tc.com"` produces three entries (and empty tokens are dropped) would lock in this behavior — currently a regression that emitted `["a.com,", "b.com\tc.com"]` would pass everything.

test/integration/test_s3.py
```diff
@@ -227,5 +228,182 @@ def test_s3_restore_directory_list():
 	# remove backup artifacts
 	for path in paths:
 		backup_to_directory(path, *COMMON_S3_OPTS, '--remove-artifacts', env=S3_ENV)
-	
-	min_srv.stop_minio_server(MINIO_NAME)
\ No newline at end of file
+
+	min_srv.stop_minio_server(MINIO_NAME)
+
+def _dead_proxy_env(extra=None):
```
dwelch-spike, line 236: From Tests agent: **Minor:** `_dead_proxy_env` binds a socket to `127.0.0.1:0`, reads the port, then closes the socket before returning. There's a small race where another process can grab that port before asbackup runs, which would flip the "with flag fails" assertions in Part 2 of `test_s3_allow_system_proxy` to false and produce a confusing CI flake. Keep the listener socket open (or use a closed-on-accept sentinel) so the negative outcome is anchored to an actively-refusing port rather than a hopefully-still-free one.

test/integration/test_s3.py
```diff
@@ -227,5 +228,182 @@ def test_s3_restore_directory_list():
...
+	# Part 2: with --s3-allow-system-proxy libcurl honors HTTP_PROXY → routes
+	# through the dead proxy → backup fails.
+	bup, path = backup_async(
+		lambda context: None,
+		env=dead_env,
+		backup_opts=COMMON_S3_OPTS + ['--remove-files', '--s3-allow-system-proxy'],
+		path=path,
+	)
+	bup_ret = lib.sync_wait(bup.wait())
+	assert bup_ret != 0, \
```
dwelch-spike: From Tests agent: **Minor:** `assert bup_ret != 0` proves the process exited non-zero, but doesn't prove the failure was a proxy connect refusal vs. some unrelated error (config parse, S3 auth path change, etc.). Since Part 2's whole purpose is to demonstrate "the proxy was actually attempted", also assert that stderr contains a proxy/connect-refused indicator (the dead port number, or `Couldn't connect`/`proxy`). Same for the restore variant at line 363.

dwelch-spike:
Takeaways from our talk. @a-spiker I think if you try the aws sdk upgrade, and that works well and allows us to get rid of all the parsing code, lets do that. If it requires that we use a new aws sdk client type, or a new library, lets not use that and instead fix the one flagged important issue in this PR.

dwelch-spike:
## Re-review after the SDK-upgrade refactor

Walked the new diff. The architectural pivot is the right call — `s3_proxy.cc`/`test_s3_proxy.c` deleted, custom parser collapsed to `conf.allowSystemProxy = s3_api.allow_system_proxy` at `src/s3_api.cc:324`, SDK bumped to 1.11.813 across `backup-build.yml`, `mac-artifact.yml`, and `install_deps.sh` (verified consistent — no straggler 1.10.55 pins anywhere). The `vaLog` virtual was correctly re-added to `AsbackupLogger` to match 1.11.x's `FormattedLogSystem`. Net effect: ~313 additions / 83 deletions, vs. ~1,251 / 11 in the original revision.

### Prior review status

| Prior finding | Status |
|---|---|
| `%00` silent credential truncation (`src/s3_proxy.cc:205`) | ✅ Resolved (file deleted) |
| Trailing-colon silent default port (`src/s3_proxy.cc:254`) | ✅ Resolved (file deleted) |
| `parse_bad_percent_encoding` test only exercises password (`test_s3_proxy.c:245`) | ✅ Resolved (file deleted) |
| NO_PROXY tokenizer has no unit test (`src/s3_api.cc:376`) | ✅ Resolved (tokenizer gone; SDK handles natively) |
| `_dead_proxy_env` bind-then-close race (`test/integration/test_s3.py:234`) | ⚠️ Still open |
| `assert bup_ret != 0` doesn't prove failure was proxy-related (`test/integration/test_s3.py:307`) | ⚠️ Still open |

### New findings from this round (4 inline)

- **Important** — `src/s3_api.cc:315`: inline `strncasecmp("https://", 8)` has zero test coverage; the deleted `s3_proxy_scheme_for_endpoint` had table-driven unit tests for case-insensitivity, prefix length, and `httpsfoo` non-match.
- **Minor** — `src/s3_api.cc:324`: no unit test calls `S3API::SetAllowSystemProxy` directly or asserts the C++ glue layer wires through to `conf.allowSystemProxy`. Only TOML round-trip is covered.
- **Minor** — `src/backup_config.c` / `src/restore_config.c` help text: doesn't warn that combining `--s3-allow-system-proxy` with `--s3-log-level=Debug`/`Trace` can echo proxy credentials (e.g. from `HTTPS_PROXY=https://user:pass@proxy:3128`) to stderr via the SDK.
- **Minor** — `.github/packaging/project/install_deps.sh`: the 9-line aws-sdk-cpp clone+build recipe is duplicated verbatim across 10 platform branches (debian11/12/13, ubuntu20/22/24, el8/9/10, amzn2023). A single helper function would prevent future drift.

### Items checked and clean

- `Aws::S3::S3Client(conf, PayloadSigningPolicy::Always, false)` overload still exists at 1.11.813 — verified.
- `vaLog` signature matches `FormattedLogSystem` in 1.11.813 — verified.
- `strncasecmp` reading 8 bytes against a possibly-7-byte `"https:/"` input is safe — POSIX guarantees stop-at-NUL.
- `allow_system_proxy` lifecycle: ctor init → setter → consumed once in `_init_api` → reset in `Shutdown`. Symmetric.
- README and all workflows show 1.11.813 consistently — no version drift.

Nothing in the new code path looks like a correctness bug.

src/s3_api.cc
```diff
@@ -298,13 +309,19 @@ S3API::_init_api(S3API& s3_api)
 	}
 	if (!s3_api.endpoint.empty()) {
 		conf.endpointOverride = s3_api.endpoint;
-		conf.scheme = Aws::Http::Scheme::HTTP;
+		// Mirror the scheme in the override URL so conf.scheme stays
+		// consistent with endpointOverride. Bare host:port defaults to HTTP.
+		conf.scheme =
+				(strncasecmp(s3_api.endpoint.c_str(), "https://", 8) == 0)
```
dwelch-spike: From Tests agent: **Important:** Inlining `strncasecmp(..., "https://", 8) == 0 ? HTTPS : HTTP` here drops the unit-test coverage the deleted `s3_proxy_scheme_for_endpoint` had (case-insensitive scheme, exact-prefix-length, `httpsfoo://...` non-match). Every integration test hits `127.0.0.1:9000` — bare host:port — so only the HTTP fallback branch is exercised in CI. Regressions that would slip through: swapping `strncasecmp` for `strncmp`, flipping the ternary, changing `8` to `7`, comparing against `"https"` instead of `"https://"` (would match `httpsfoo`). Either pull this back into a tiny static helper with a libcheck table-test, or replace it with `Aws::Http::URI(endpoint).GetScheme()` so the SDK owns it.

src/s3_api.cc
```diff
@@ -298,13 +309,19 @@ S3API::_init_api(S3API& s3_api)
...
+						? Aws::Http::Scheme::HTTPS
+						: Aws::Http::Scheme::HTTP;
 	}
 
 	conf.maxConnections = std::max(s3_api.max_async_downloads,
 			s3_api.max_async_uploads);
-	
+
 	conf.connectTimeoutMs = s3_api.connect_timeout_ms;
+	conf.allowSystemProxy = s3_api.allow_system_proxy;
```
dwelch-spike, line 335: From Tests agent: **Minor:** The only test coverage for the C++ wiring is the TOML round-trip in `test_backup_conf.c` / `test_restore_conf.c`, which asserts `s3-allow-system-proxy=true` populates the C `backup_config_t::s3_allow_system_proxy` field — but nothing asserts that calling `S3API::SetAllowSystemProxy(true)` actually causes this line (`conf.allowSystemProxy = s3_api.allow_system_proxy`) to fire. A future refactor that accidentally drops this assignment is only caught by the integration test, which has its own weak failure-mode discrimination (see comment on `test_s3.py:307`). A small libcheck test that constructs an `S3API`, calls `SetAllowSystemProxy(true)`, and reads the field back closes the gap without needing MinIO.

src/backup_config.c
```diff
@@ -1549,6 +1556,13 @@ usage(const char *name)
...
+	fprintf(stdout, "      --s3-allow-system-proxy\n");
+	fprintf(stdout, "                      Allow the S3 client to honor the standard system proxy\n");
+	fprintf(stdout, "                      environment variables (https_proxy / HTTPS_PROXY,\n");
+	fprintf(stdout, "                      http_proxy, all_proxy / ALL_PROXY, no_proxy / NO_PROXY)\n");
+	fprintf(stdout, "                      when contacting S3. Precedence and supported proxy URL\n");
+	fprintf(stdout, "                      syntax follow libcurl. By default these variables are\n");
+	fprintf(stdout, "                      ignored.\n\n");
 
 	fprintf(stdout, "\n\n");
 	fprintf(stdout, "Default configuration files are read from the following files in the given order:\n");
```
dwelch-spike, line 1572: From Security agent: **Minor:** When `--s3-allow-system-proxy` is enabled, the SDK passes through to libcurl/aws-c-http, both of which log the full configured proxy URL at `Debug`/`Trace` verbosity. Operators commonly embed credentials in the proxy URL (`HTTPS_PROXY=https://user:pass@proxy.corp:3128`), so combining this flag with `--s3-log-level=Debug` or `--s3-log-level=Trace` will write those credentials to stderr (and downstream to systemd-journald, CI logs, log aggregators). The flag is opt-in and the default is safe, but the help text should warn operators to either scrub credentials from proxy URLs before raising verbosity or use a credential-free proxy. Mirror the same note in `src/restore_config.c`.

.github/packaging/project/install_deps.sh
```diff
@@ -77,10 +77,9 @@ function compile_deps_debian11() {
...
 
-  git clone https://github.com/aws/aws-sdk-cpp.git
+  git clone --depth 1 --branch $AWS_SDK_VERSION https://github.com/aws/aws-sdk-cpp.git
   pushd aws-sdk-cpp
-  git checkout $AWS_SDK_VERSION
-  git submodule update --init --recursive
+  git submodule update --init --recursive --depth 1
   mkdir build
   cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_ONLY="s3" -DBUILD_SHARED_LIBS=OFF -DENABLE_TESTING=OFF -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_INSTALL_LIBDIR=lib -DENABLE_UNITY_BUILD=ON
   make_parallel -C build
```
dwelch-spike: From Architecture agent: **Minor:** This 9-line aws-sdk-cpp clone+build recipe (clone, submodule init, mkdir, cmake with identical `-D` flags, `make_parallel`, install) is now copy-pasted verbatim across 10 platform branches: debian11/12/13, ubuntu20/22/24, el8/9/10, amzn2023. The cleanup in this PR (added `--depth 1 --branch`, dropped `git checkout`, removed s2n-tls — good) had to be applied in 10 places. Any future SDK option change (`BUILD_ONLY`, `ENABLE_UNITY_BUILD`, a cmake policy pin) will need the same 10-touch update, and CI won't catch drift between branches. A single `build_aws_sdk_cpp()` helper alongside the existing `build_libyaml_static()` would collapse the duplication. Out of scope for this PR is defensible — but worth a follow-up ticket while the file is open.

dwelch-spike:
## Re-review after `fix: address post-refactor review findings`

All four new findings from the prior round are addressed, plus both lingering integration-test findings (one fully, one with improved message). Subagent sweep (Logic, Control Flow, Data Flow, Tests, Architecture) turned up no new correctness defects.

### Prior findings — status

| Finding | Status |
|---|---|
| Inline `strncasecmp` had no test coverage (`src/s3_api.cc:315`) | ✅ Extracted as `S3API::SchemeForEndpoint` + 8 table-driven tests in `test/unit/test_s3_config.c` including the `httpsfoo` boundary case |
| No unit test for C++ glue (`src/s3_api.cc:324`) | ✅ Added `S3API::GetAllowSystemProxy` + `s3_get_allow_system_proxy` C wrapper + 3 round-trip tests |
| Credential-leak warning missing from help text | ✅ Added 4-line note to `backup_config.c` and `restore_config.c` usage strings, calls out HTTPS_PROXY/HTTP_PROXY credentials and `--s3-log-level Debug`/`Trace` |
| `install_deps.sh` SDK build recipe duplicated 10× | ✅ Replaced with `build_aws_sdk_cpp()` helper; pushd/popd stack balance verified across all callers |
| `_dead_proxy_env` bind-then-close race | ✅ Converted to `@contextlib.contextmanager` that holds the socket open across the entire `with` block; explicit comment about deliberately not calling `listen()` so connections get `ECONNREFUSED` |
| `assert bup_ret != 0` doesn't prove proxy-related failure | ⚠️ Message improved to mention `ECONNREFUSED`, assertion itself unchanged. Part 1 of each test catches gross binary/option failures, so the practical risk is bounded — but a future regression that breaks asbackup *after* getopt parses successfully would still pass. Noted as still-open for a follow-up; not blocking |

### New findings from this round

No correctness defects. Architecture agent flagged the test-seam additions:

- `s3_scheme_for_endpoint_is_https` and `s3_get_allow_system_proxy` in `include/file_proxy.h` and `S3API::SchemeForEndpoint` in `include/s3_api.h` widen the production-facing API surface purely to give the unit test a callable seam. Cleaner long-term: an internal test-only header (`test/unit/s3_testing.h`) with the forward declarations, or anonymous-namespace free functions in `s3_api.cc` that the test calls directly. Today this is contained to two predicate-shaped functions, but it sets a precedent for "add to the public header when a test needs it."
- `s3_scheme_for_endpoint_is_https` collapses `Aws::Http::Scheme` (a 2-value enum today, but the SDK reserves the right to grow it) into a single bool. If a future SDK adds `Scheme::WS` or similar, callers reading `false` will silently treat it as HTTP. Not load-bearing today.

Both are minor enough that I would not block on them — file as follow-up tickets and move on.

### Items checked clean
- `build_aws_sdk_cpp()` pushd/popd balance: helper is net-zero, all callers (debian/ubuntu cwd-based, EL+amzn2023 wrapped in explicit `pushd /opt` / `popd`) are balanced — including amzn2023, which correctly has the trailing `popd`
- `_dead_proxy_env` exception safety: the inner `with socket.socket()` ensures the socket closes cleanly even if a test inside the with-block raises
- Test isolation: `test_allow_system_proxy_set_true` does restore the default after asserting, and `set_false` is order-independent so the leak-on-assertion-failure risk is bounded
- `s3_config_suite()` is correctly wired into `backup_tests.h` and `main.c`; Makefile globs pick it up
- `SchemeForEndpoint` taking `std::string` by const-ref while the C shim does `std::string(endpoint)` after a NULL guard — no NULL deref reachable
- 8 scheme tests cover the boundary that matters (`httpsfoo` non-match), case-insensitive variants, NULL/empty

LGTM from a correctness standpoint. Ship it.

dwelch-spike:
> ## Re-review after `fix: address post-refactor review findings`
> 
> All four new findings from the prior round are addressed, plus both lingering integration-test findings (one fully, one with improved message). Subagent sweep (Logic, Control Flow, Data Flow, Tests, Architecture) turned up no new correctness defects.
> ### Prior findings — status
> Finding 	Status
> Inline `strncasecmp` had no test coverage (`src/s3_api.cc:315`) 	✅ Extracted as `S3API::SchemeForEndpoint` + 8 table-driven tests in `test/unit/test_s3_config.c` including the `httpsfoo` boundary case
> No unit test for C++ glue (`src/s3_api.cc:324`) 	✅ Added `S3API::GetAllowSystemProxy` + `s3_get_allow_system_proxy` C wrapper + 3 round-trip tests
> Credential-leak warning missing from help text 	✅ Added 4-line note to `backup_config.c` and `restore_config.c` usage strings, calls out HTTPS_PROXY/HTTP_PROXY credentials and `--s3-log-level Debug`/`Trace`
> `install_deps.sh` SDK build recipe duplicated 10× 	✅ Replaced with `build_aws_sdk_cpp()` helper; pushd/popd stack balance verified across all callers
> `_dead_proxy_env` bind-then-close race 	✅ Converted to `@contextlib.contextmanager` that holds the socket open across the entire `with` block; explicit comment about deliberately not calling `listen()` so connections get `ECONNREFUSED`
> `assert bup_ret != 0` doesn't prove proxy-related failure 	⚠️ Message improved to mention `ECONNREFUSED`, assertion itself unchanged. Part 1 of each test catches gross binary/option failures, so the practical risk is bounded — but a future regression that breaks asbackup _after_ getopt parses successfully would still pass. Noted as still-open for a follow-up; not blocking
> ### New findings from this round
> 
> No correctness defects. Architecture agent flagged the test-seam additions:
> 
>     * `s3_scheme_for_endpoint_is_https` and `s3_get_allow_system_proxy` in `include/file_proxy.h` and `S3API::SchemeForEndpoint` in `include/s3_api.h` widen the production-facing API surface purely to give the unit test a callable seam. Cleaner long-term: an internal test-only header (`test/unit/s3_testing.h`) with the forward declarations, or anonymous-namespace free functions in `s3_api.cc` that the test calls directly. Today this is contained to two predicate-shaped functions, but it sets a precedent for "add to the public header when a test needs it."
> 
>     * `s3_scheme_for_endpoint_is_https` collapses `Aws::Http::Scheme` (a 2-value enum today, but the SDK reserves the right to grow it) into a single bool. If a future SDK adds `Scheme::WS` or similar, callers reading `false` will silently treat it as HTTP. Not load-bearing today.
> 
> 
> Both are minor enough that I would not block on them — file as follow-up tickets and move on.
> ### Items checked clean
> 
>     * `build_aws_sdk_cpp()` pushd/popd balance: helper is net-zero, all callers (debian/ubuntu cwd-based, EL+amzn2023 wrapped in explicit `pushd /opt` / `popd`) are balanced — including amzn2023, which correctly has the trailing `popd`
> 
>     * `_dead_proxy_env` exception safety: the inner `with socket.socket()` ensures the socket closes cleanly even if a test inside the with-block raises
> 
>     * Test isolation: `test_allow_system_proxy_set_true` does restore the default after asserting, and `set_false` is order-independent so the leak-on-assertion-failure risk is bounded
> 
>     * `s3_config_suite()` is correctly wired into `backup_tests.h` and `main.c`; Makefile globs pick it up
> 
>     * `SchemeForEndpoint` taking `std::string` by const-ref while the C shim does `std::string(endpoint)` after a NULL guard — no NULL deref reachable
> 
>     * 8 scheme tests cover the boundary that matters (`httpsfoo` non-match), case-insensitive variants, NULL/empty
> 
> 
> LGTM from a correctness standpoint. Ship it.

I don't think the outstanding test issue is much of a problem, this project will be replaced before long.

dwelch-spike (approved):
As long as the tests pass this LGTM

PR #141, opened 2026-05-27
ci: build and test Docker images on dry-run, push only on release

a-spiker:
## Summary

- Removes the job-level `if: inputs.release == true` guard from `docker-build-arch` so Docker images are **always** built and smoke-tested, even during dry-run dispatches (`release=false`)
- Moves the publish gate (`if: inputs.release == true`) down to the three push-only steps: **Install JFrog CLI**, **Docker login to Artifactory**, and **Push native per-arch tag to JFrog**
- Adds `if: inputs.release == true` to the `docker-manifest` job (pure push operation — no change to behaviour when `release=true`)

## Behaviour after this change

| Step | `release=false` (dry-run) | `release=true` (publish) |
|---|---|---|
| Docker build + smoke test | ✅ now runs | ✅ |
| JFrog login / push per-arch tag | ❌ skipped | ✅ |
| Docker multi-arch manifest | ❌ skipped | ✅ |

## Test plan
- [ ] Dispatch with `release=false` and confirm docker-build-arch runs build/test steps but skips JFrog login and push
- [ ] Dispatch with `release=true` and confirm full publish flow is unchanged

### DIFF
--- a/.github/bin/install_deps.sh
+++ b/.github/bin/install_deps.sh
@@ -341,12 +341,14 @@ function compile_deps_el8() {
   make install
 
   pushd /opt
-  git clone https://git.savannah.gnu.org/git/readline.git
-  pushd readline
-  git checkout readline-8.3
+  curl -LO https://mirrors.ocf.berkeley.edu/gnu/readline/readline-8.3.tar.gz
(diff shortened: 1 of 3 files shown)


pvinh-spike (approved):
LGTM

PR #139, opened 2026-05-27
TOOLS-3599: Ubuntu 26.04 CI and packaging fixes

pvinh-spike:
## Summary
- Add **Ubuntu 26.04** to **Build and Release** and **build-check** (pinned `ubuntu:resolute-20260413`), plus **install_deps** / **build_package** support (including **libjitterentropy3-dev** and gated **`-ljitterentropy`** for Ubuntu 26 static OpenSSL).
- Fix **macOS** CI issues: **Homebrew `brew info cmake` / JSON**, **AWS SDK 1.11.813** on release mac, **AWS SDK** compile on **macOS 26**, and **static-binary sanity** checks that are compatible with **`pipefail`**.
- Fix **Linux** packaging/runtime: **OpenSSL link order** on EL vs `/usr/local`, **deb** dependency **`libjitterentropy3`** for Ubuntu 26, and **`dnf update`** in **RPM install tests** so OpenSSL symbol versions match packaging builds.

## Test plan
- [x] **Build and Release** workflow green on this branch.
- [ ] Spot-check **install test** matrix (especially **el9** / **ubuntu26.04**) on the merged result if anything changes post-review.

PR #149, opened 2026-06-12
fix: build AWS SDK on macos-26 under Xcode 26.5

a-spiker:
## Problem

The `Mac Artifact` workflow is failing on `master`. The `macos-26` job is the real failure; `macos-14` / `macos-15-intel` are just canceled by matrix `fail-fast`.

The macos-26 runner image moved to **Xcode 26.5**. The same workflow passed on May 29 (commit #143) and #144 ("Drop support for ubuntu-20.04") changed nothing in the Mac build, so this is an environment regression, not a code change.

AWS SDK C++ `1.11.813` no longer builds under Xcode 26.5 with the current config:

1. The cached AWS SDK cmake `build_static` tree was configured against the previous Xcode SDK. `make install` rebuilds it under the new toolchain, producing libc++ header search-path errors (`<cctype> tried including <ctype.h> but didn't find libc++'s <ctype.h>` ... 11 of these).
2. Xcode 26.5's clang flags `abs(int64_t)` in `AdaptiveRetryStrategy.cpp:80` as `-Werror,-Wabsolute-value`.

The existing `-Wno-deprecated-literal-operator` workaround (#139) only covers an older, separate warning.

## Fix

- Add the Xcode version to the AWS SDK cache key. A toolchain bump now invalidates the stale cache and forces a clean reconfigure + rebuild against the current SDK, which clears the header search-path errors. Also prevents this class of breakage recurring on future image bumps.
- Add `-Wno-absolute-value` alongside the existing `-Wno-deprecated-literal-operator` for the macos-26 SDK build.

## Validation

Draft, pushed to let CI exercise the macos-26 job. If Xcode 26.5 surfaces further `-Werror` warnings beyond these two, we add the matching `-Wno-*` flags.

### DIFF
--- a/.github/workflows/mac-artifact.yml
+++ b/.github/workflows/mac-artifact.yml
@@ -109,6 +109,10 @@ jobs:
         # Avoid `brew info | grep -m 1` (broken pipe). Homebrew JSON may be an array or one object.
         CMAKE_VERSION="$(brew info cmake --json=v2 | jq -r '(if type == "array" then .[0] else . end).versions.stable' | sed -E 's/^([0-9]+\.[0-9]+\.[0-9]+).*/\1/')"
         echo "CMAKE_VERSION=$CMAKE_VERSION" >> $GITHUB_ENV
+        # Key caches on the Xcode version too: a runner-image Xcode bump (e.g. 26.4 -> 26.5)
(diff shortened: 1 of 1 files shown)


pvinh-spike (approved):
LGTM

PR #154, opened 2026-06-18
ci: build AWS SDK on macos-26 under Xcode 26.5 in build-and-release

a-spiker:
build-and-release.yml carries its own inline copy of the macOS AWS SDK build. The b54c999 fix was only applied to mac-artifact.yml, so the release workflow still broke on the macos-26 runner once its image moved to Xcode 26.5 / AppleClang 21:

- The AWS SDK cache key had no Xcode component, so the Xcode 26.4.1 -> 26.5 image bump reused a stale build_static tree. `sudo make install` then re-ran cmake against absolute Xcode_26.4.1 SDK paths under the 26.5 compiler, producing libc++ header-search-path errors (<cwctype>/<wctype.h> "not configured properly").
- That recompile lacked -Wno-absolute-value, so std::abs(int64_t) in AdaptiveRetryStrategy.cpp tripped -Werror,-Wabsolute-value.

Port the mac-artifact.yml fix: key the cache on XCODE_VERSION (busts stale trees on a toolchain bump) and suppress both -Wabsolute-value and -Wdeprecated-literal-operator for macos-26.


dwelch-spike (approved):
LGTM

PR #156, opened 2026-06-19
ci: publish release bundles under direct VERSION (uniform across tools)

a-spiker:
## What

Drop the branch-prefixed `BUNDLE_VERSION` and publish every JFrog release bundle under the SemVer `VERSION` directly, matching `aerospike-admin` (asadm). This makes the release-bundle version scheme uniform across all five tools.

## Why

Non-release / feature-branch publishes were doubling the branch name, e.g. `master-4.5.6-master.<sha>`, because `VERSION` already embeds `<branch>.<sha>` as a SemVer pre-release tag and `BUNDLE_VERSION` prepended `<branch>-` again. The SHA already disambiguates, so the prefix was redundant.

## Resulting version pattern (now identical for all 5 tools)

- Real release (master, not skip_tagging): `4.5.6`
- Non-release / feature / skip_tagging: `4.5.6-master.<sha>` (branch appears once)

## Note

This branch also carries earlier related CI commit(s) that were already on it. Draft for review.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -107,7 +107,6 @@ jobs:
       contents: read
     outputs:
       VERSION: ${{ steps.compute.outputs.version }}
-      BUNDLE_VERSION: ${{ steps.compute.outputs.bundle_version }}
     steps:
       - name: Compute VERSION (SemVer-aware for branch builds)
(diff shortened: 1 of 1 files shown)


pvinh-spike (approved):
LGTM

PR #157, opened 2026-06-23
ci: build EL9/EL10 against oldest-stable OpenSSL to fix dynamic linking

a-spiker:
## Problem

On RHEL 9 / RHEL 10, `asbackup`/`asrestore` failed:

```
/usr/bin/asbackup: /lib64/libcrypto.so.3: version `OPENSSL_3.4.0' not found (required by /usr/bin/asbackup)
asbackup: symbol lookup error: asbackup: undefined symbol: EVP_MD_CTX_get_size_ex, version OPENSSL_3.4.0   # at S3 init
```

`asbackup` links `libcrypto`/`libssl` **dynamically** on EL. The build ran `dnf -y update`, which floated OpenSSL to 3.5 (RHEL rebased 3.0 → 3.5 in 9.7 / 10.1), so the AWS SDK's `aws-c-cal` recorded `OPENSSL_3.4.0` symbol dependencies (`EVP_MD_CTX_get_size_ex`) that the older system `libcrypto` on RHEL 9.x / 10.0 can't satisfy. The test job hid the regression by also running `dnf -y update`.

## Fix

Keep the bindings dynamic, but build against the **oldest-stable (GA) OpenSSL** so the symbol floor is the lowest the major shipped.

This pin is **new to the tools suite**, not an existing convention. `aql`/`asadm` reference the Rocky vault only to pull individual RPMs by URL, and they run `dnf -y update` (floating OpenSSL up rather than pinning it). The reason asbackup needs the pin and they don't is asbackup-specific: it statically links the AWS SDK (`aws-c-cal`), which binds newer `OPENSSL_3.4.0` symbols that `aql`/`asadm` never pull in. What this PR does share with the siblings is the UBI base images (`ubi9:9.6`, `ubi10:10.0`) and the 9.6 / 10.0 minor pairing.

- New `.github/bin/pin_build_repos.sh` freezes EL9/EL10 to the Rocky vault GA snapshot matching the UBI base (9.6 / 10.0) → GA OpenSSL **3.2.2**, floor `OPENSSL_3.2.0`. Package signatures are verified against the Rocky release key fetched over HTTPS (`gpgcheck=1`), since UBI images don't ship Rocky's keyring. No-op on other distros.
- The OpenSSL `distro-sync` fails loudly (no `2>/dev/null || true`): a blocked or failed downgrade stops the job instead of silently leaving a refloated OpenSSL and a falsely-green backstop.
- Removed `dnf -y update` from `install_deps_el9` / `install_deps_el10` and from the install-test prereqs, so the test runs against the GA `libcrypto` real customers have.
- New `.github/bin/assert_openssl_floor.sh` is a container-independent backstop: it inspects the built binary's dynamic symbol table (`objdump -T`) and fails the build if any referenced OpenSSL symbol version exceeds `OPENSSL_3.2.0`. This catches lazily bound symbols that `asbackup --help` never exercises. Wired into the build job (EL9/EL10) and `build-check` (EL10 per-commit).

Result: binaries require ≤ `OPENSSL_3.2.0` and load on all RHEL 9.4+ / 10.x.

## Testing

- [ ] Dispatch **Build and Release** with `release=false` (dry-run) on this branch and confirm the el9/el10 build + install-test pass on amd64 and arm64.
- The `assert_openssl_floor.sh` step enforces the symbol floor at build time, so a refloat fails CI directly rather than only at a customer's RHEL.

## Notes

- Floor is `OPENSSL_3.2.0` (covers RHEL 9.4+ and all 10.x). RHEL **9.0-9.3** (OpenSSL 3.0) would require pinning el9 to the Rocky vault **9.0** snapshot, which forces a downgrade on the 9.6 base or a 9.0 base image. Not included here; can add if needed.
- `el8` (OpenSSL 1.1.1, different SONAME) and `amzn2023` are unchanged.
- Per-commit `build-check` covers EL10 (not EL9); EL9 is exercised by the full Build and Release workflow.

### DIFF
--- a/.github/bin/pin_build_repos.sh
+++ b/.github/bin/pin_build_repos.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+# Freeze EL9/EL10 to the Rocky vault GA snapshot for the base image's minor, so the
+# dynamically linked OpenSSL floor stays at the oldest the major shipped (3.2.2 ->
+# OPENSSL_3.2.0). The default UBI repos float to 3.5 (9.7/10.1), whose OPENSSL_3.4.0
+# symbols (e.g. EVP_MD_CTX_get_size_ex via the AWS SDK) the older RHEL 9.x/10.0
+# libcrypto can't satisfy. Run as a subprocess (not sourced); no-op on other distros.
+set -euo pipefail
+
+distro="${1:-${ENV_DISTRO:-}}"
+case "$distro" in
+  el9)  rel=9.6  ;;   # match the ubi9:9.6 / ubi10:10.0 base GA (no downgrade)
+  el10) rel=10.0 ;;
+  *)    echo "pin_build_repos: nothing to pin for '${distro:-<unset>}'"; exit 0 ;;
+esac
+
+arch="$(uname -m)"
+base="https://dl.rockylinux.org/vault/rocky/${rel}"
+
+# Replace the floating UBI repos with the frozen vault snapshot. gpgcheck off: official
+# Rocky HTTPS mirrors, and the Rocky keys aren't in a UBI image (as aql/asadm do).
+rm -f /etc/yum.repos.d/*.repo
+: > /etc/yum.repos.d/rocky-vault.repo
+for repo in BaseOS AppStream CRB devel; do
+  name="$(printf '%s' "$repo" | tr '[:upper:]' '[:lower:]')"
+  {
+    printf '[%s]\n' "$name"
+    printf 'name=Rocky Linux %s - %s (vault)\n' "$rel" "$repo"
+    printf 'baseurl=%s/%s/%s/os/\n' "$base" "$repo" "$arch"
+    printf 'gpgcheck=0\n'
+    printf 'enabled=1\n\n'
+  } >> /etc/yum.repos.d/rocky-vault.repo
+done
+
+# Pull OpenSSL back down to the snapshot in case a prior step floated it.
+dnf -y distro-sync openssl openssl-libs 2>/dev/null || true
+
+echo "pin_build_repos: ${distro} pinned to Rocky vault ${rel} (${arch})"
--- a/.github/bin/pin_build_repos.sh
+++ b/.github/bin/pin_build_repos.sh
@@ -35,13 +35,28 @@ for repo in BaseOS AppStream CRB devel; do
   } >> /etc/yum.repos.d/rocky-vault.repo
 done
 
-# Pull OpenSSL back down to the snapshot in case an earlier "Install prereqs" step
-# floated it forward against the still-default UBI repos. This is the only thing that
-# re-establishes the GA floor in the test job, so let it fail loudly (no `2>/dev/null
-# || true`): a silently swallowed sync would leave a refloated OpenSSL in place and
-# turn the regression backstop green against the wrong libcrypto. The post-sync rpm -q
-# records the resolved version so the floor is visible in the job log.
-dnf -y distro-sync openssl openssl-libs
-rpm -q --qf 'pin_build_repos: synced %{NAME} to %{VERSION}-%{RELEASE}\n' openssl-libs
+# Pull OpenSSL down to the GA floor (3.2.x) only if an earlier step floated it past it.
+# A plain distro-sync would force a same-version reinstall even when nothing floated,
+# which on el10 fails on a file conflict: the vault's openssl-libs bundles fips.so, but
+# the ubi10 base splits it into openssl-fips-provider-so. Acting only on a real float
+# avoids that, and the post-check keeps the pin self-verifying: it fails loudly (no
+# `2>/dev/null || true`) rather than leaving a refloated OpenSSL the backstop can't see.
+floor_minor=2
+ver="$(rpm -q --qf '%{VERSION}' openssl-libs)"
+case "$ver" in
+  3.*) minor="${ver#3.}"; minor="${minor%%.*}" ;;
+  *)   echo "pin_build_repos: unexpected openssl-libs version '${ver}'" >&2; exit 1 ;;
+esac
+if [ "$minor" -gt "$floor_minor" ]; then
+  echo "pin_build_repos: openssl-libs ${ver} floated above 3.${floor_minor}; syncing down to the vault"
+  dnf -y distro-sync openssl openssl-libs
+  ver="$(rpm -q --qf '%{VERSION}' openssl-libs)"
+  minor="${ver#3.}"; minor="${minor%%.*}"
+fi
+if [ "$minor" -gt "$floor_minor" ]; then
+  echo "pin_build_repos: openssl-libs still ${ver}, above the 3.${floor_minor} floor" >&2
+  exit 1
+fi
+echo "pin_build_repos: openssl-libs at ${ver} (within the 3.${floor_minor} floor)"
 
 echo "pin_build_repos: ${distro} pinned to Rocky vault ${rel} (${arch})"
--- a/.github/bin/pin_build_repos.sh
+++ b/.github/bin/pin_build_repos.sh
@@ -8,16 +8,19 @@ set -euo pipefail
 
 distro="${1:-${ENV_DISTRO:-}}"
 case "$distro" in
-  el9)  rel=9.6  ;;   # match the ubi9:9.6 / ubi10:10.0 base GA (no downgrade)
-  el10) rel=10.0 ;;
+  el9)  rel=9.6  ;;   # match the ubi9:9.6 base GA (openssl-libs 3.2.2, no downgrade)
+  el10) rel=10.0 ;;   # match the ubi10:10.0 base GA (openssl-libs 3.2.2, no downgrade)
   *)    echo "pin_build_repos: nothing to pin for '${distro:-<unset>}'"; exit 0 ;;
 esac
 
 arch="$(uname -m)"
 base="https://dl.rockylinux.org/vault/rocky/${rel}"
 
-# Replace the floating UBI repos with the frozen vault snapshot. gpgcheck off: official
-# Rocky HTTPS mirrors, and the Rocky keys aren't in a UBI image (as aql/asadm do).
+# Replace the floating UBI repos with the frozen vault snapshot. Keep signature
+# verification on: UBI images don't ship Rocky's keyring, so point dnf at the Rocky
+# release key over HTTPS and let it import on first use. gpgcheck=1 guards against a
+# poisoned mirror/CDN swapping the libcrypto/libssl these binaries link against; HTTPS
+# only authenticates the transport to dl.rockylinux.org, not the package contents.
 rm -f /etc/yum.repos.d/*.repo
 : > /etc/yum.repos.d/rocky-vault.repo
 for repo in BaseOS AppStream CRB devel; do
@@ -26,12 +29,19 @@ for repo in BaseOS AppStream CRB devel; do
     printf '[%s]\n' "$name"
     printf 'name=Rocky Linux %s - %s (vault)\n' "$rel" "$repo"
     printf 'baseurl=%s/%s/%s/os/\n' "$base" "$repo" "$arch"
-    printf 'gpgcheck=0\n'
+    printf 'gpgcheck=1\n'
+    printf 'gpgkey=https://dl.rockylinux.org/pub/rocky/RPM-GPG-KEY-Rocky-%s\n' "${rel%%.*}"
     printf 'enabled=1\n\n'
   } >> /etc/yum.repos.d/rocky-vault.repo
 done
 
-# Pull OpenSSL back down to the snapshot in case a prior step floated it.
-dnf -y distro-sync openssl openssl-libs 2>/dev/null || true
+# Pull OpenSSL back down to the snapshot in case an earlier "Install prereqs" step
+# floated it forward against the still-default UBI repos. This is the only thing that
+# re-establishes the GA floor in the test job, so let it fail loudly (no `2>/dev/null
+# || true`): a silently swallowed sync would leave a refloated OpenSSL in place and
+# turn the regression backstop green against the wrong libcrypto. The post-sync rpm -q
+# records the resolved version so the floor is visible in the job log.
+dnf -y distro-sync openssl openssl-libs
+rpm -q --qf 'pin_build_repos: synced %{NAME} to %{VERSION}-%{RELEASE}\n' openssl-libs
 
 echo "pin_build_repos: ${distro} pinned to Rocky vault ${rel} (${arch})"
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -192,6 +192,10 @@ jobs:
           fetch-tags: true
           submodules: recursive
 
+      # EL9/EL10: build against the oldest-stable (GA) OpenSSL; no-op elsewhere.
+      - name: Pin EL build repos to frozen vault
+        run: bash .github/bin/pin_build_repos.sh "${{ matrix.distro }}"
+
       - name: Install build dependencies
         env:
           ENV_DISTRO: ${{ matrix.distro }}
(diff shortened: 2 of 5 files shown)


dwelch-spike:
**PR Review (version 2026.05.27)**

**Promises:**
1. `asbackup`/`asrestore` on EL link `libcrypto`/`libssl` dynamically and must load on RHEL 9.4+ / all 10.x (symbol floor ≤ `OPENSSL_3.2.0`).
2. EL9/EL10 builds freeze to a Rocky vault GA snapshot (9.6/10.0) so OpenSSL stays at GA 3.2.2 instead of floating to 3.5.
3. The masking `dnf -y update` is removed from the build deps and the test job, so the test runs against the GA `libcrypto` real customers have (regression backstop).
4. No-op on non-EL distros (el8 / amzn2023 / ubuntu / debian / macOS unchanged).
5. The approach "follows the suite convention (UBI base + Rocky vault, same as aql/asadm)".

| Gate | Status |
|------|--------|
| Contract | Fail — promise 5 unsupported; promises 1–2 lack independent evidence |
| Impacted surface | Blind spot — el9 pin path has no per-commit coverage; pin minor can drift from base image |
| Failure paths | Fail — the pin's OpenSSL hold fails open (`2>/dev/null \|\| true`) |
| Evidence | Fail — no symbol-floor assertion; install test is `--help`/`--version` only |
| Quality | Fail — `(as aql/asadm do)` comment is inaccurate; el9 comment references ubi10 |

**4 findings** posted inline (0 Critical, 3 Important, 1 Minor).

---

**Cross-repo / recent-release consistency (the specific ask):** The mechanism here — replacing the dnf repo set with a frozen Rocky vault snapshot, `gpgcheck=0`, and `dnf distro-sync` to hold OpenSSL down — is **new to the tools suite, not a convention `aql`/`asadm` follow.** Verified against the sibling repos at their default branches:

- `aql/.github/bin/install_deps.sh` (el9 L120, el10 L131) and `aerospike-admin/.github/bin/install_deps.sh` (el9/el10 L99/L105) both **run `dnf -y update`** — they float OpenSSL *up*, the opposite of this PR. They reference the Rocky vault **only** to `yum install` individual RPMs by direct URL (`flex`, `readline-devel`); neither writes a repo file, sets `gpgcheck`, or runs `distro-sync`.
- The only things that genuinely match the siblings: the **UBI base images** (`ubi9:9.6`, `ubi10:10.0`) and the el9→9.6 / el10→10.0 minor pairing.
- `aql` is also C and also links OpenSSL **dynamically** on EL, so it has the same symbol-floor exposure — yet it does not pin. The reason this PR needs the pin and the siblings don't is asbackup-specific: the **AWS SDK (`aws-c-cal`)** binds the newer `OPENSSL_3.4.0` symbols, which `aql`/`asadm` never pull in.
- Within asbackup itself, `master` (pre-PR) also used `dnf -y update` for el9/el10, so **recent releases floated OpenSSL the same way** — this PR is the first divergence (and the regression it fixes was caused by that floating + RHEL's 3.0→3.5 rebase in 9.7/10.1).

Net: the *direction* of the fix (pin down to the GA floor) is sound and arguably more correct than the siblings — but the PR description and the in-code `(as aql/asadm do)` comment **overstate consistency**. Recommend rewording to: "asbackup links the AWS SDK, which binds newer OpenSSL symbols than aql/asadm need; this pin is new to the suite," and consider whether `aql` (same dynamic-link exposure if it ever adds an OpenSSL-3.4 consumer) should be aligned.

**Also (Minor, not blocking):** `build-check.yml`'s matrix is `ubuntu24.04 / ubuntu26.04 / el10` — no **el9**, so the `el9)` branch of `pin_build_repos.sh` is exercised only by the full release workflow, never on per-commit checks.

**Verdict:** Request changes — none are functional blockers (CI is green and the approach is reasonable), but the fix fails open on a pin/sync failure, has no independent proof the symbol floor is actually held, and ships comments/description that misrepresent the sibling convention.

.github/bin/pin_build_repos.sh
```diff
@@ -0,0 +1,37 @@
...
+  el9)  rel=9.6  ;;   # match the ubi9:9.6 / ubi10:10.0 base GA (no downgrade)
+  el10) rel=10.0 ;;
+  *)    echo "pin_build_repos: nothing to pin for '${distro:-<unset>}'"; exit 0 ;;
+esac
+
+arch="$(uname -m)"
+base="https://dl.rockylinux.org/vault/rocky/${rel}"
+
+# Replace the floating UBI repos with the frozen vault snapshot. gpgcheck off: official
+# Rocky HTTPS mirrors, and the Rocky keys aren't in a UBI image (as aql/asadm do).
```
dwelch-spike: From Architecture agent: **Important:** The `(as aql/asadm do)` attribution is inaccurate, and `gpgcheck=0` is a security tradeoff with no sibling precedent.

Direct inspection of the sibling repos: `aql/.github/bin/install_deps.sh` and `aerospike-admin/.github/bin/install_deps.sh` **never write a repo file and never set `gpgcheck`** — they `dnf -y update` and pull individual RPMs (`flex`, `readline-devel`) from the vault by direct URL. So there is no "gpgcheck off ... as aql/asadm do" precedent; the wholesale repo replacement + disabled signature verification is unique to this PR. A wrong comment on the single most security-relevant line here is worse than no comment — it tells the next maintainer this is vetted house style when it isn't.

Separately, `gpgcheck=0` means dnf installs whatever the mirror serves without verifying the Rocky signing key. HTTPS authenticates the transport to `dl.rockylinux.org`; it does **not** defend against a poisoned/compromised mirror or CDN edge — which is exactly what package signing covers. These packages become the `libcrypto`/`libssl` your shipped binaries link against, so the trust boundary is real.

Fix: import the Rocky release key over HTTPS and keep `gpgcheck=1`:
```
printf 'gpgcheck=1\n'
printf 'gpgkey=https://dl.rockylinux.org/pub/rocky/RPM-GPG-KEY-Rocky-%s\n' "${rel%%.*}"
```
If you keep `gpgcheck=0`, drop the false `(as aql/asadm do)` clause and document it as an accepted build-container risk on its own terms — not as if HTTPS made it equivalent to signing.

.github/bin/pin_build_repos.sh
```diff
@@ -0,0 +1,37 @@
...
+    printf '[%s]\n' "$name"
+    printf 'name=Rocky Linux %s - %s (vault)\n' "$rel" "$repo"
+    printf 'baseurl=%s/%s/%s/os/\n' "$base" "$repo" "$arch"
+    printf 'gpgcheck=0\n'
+    printf 'enabled=1\n\n'
+  } >> /etc/yum.repos.d/rocky-vault.repo
+done
+
+# Pull OpenSSL back down to the snapshot in case a prior step floated it.
+dnf -y distro-sync openssl openssl-libs 2>/dev/null || true
```
dwelch-spike: From Control Flow agent: **Important:** This is the one step that enforces the GA OpenSSL floor, and it is written to fail open. `2>/dev/null || true` discards both the diagnostics and the exit code, so every failure mode — a 404 on the vault `baseurl`, a transient mirror/network error, a dnf solve error, or a downgrade conflict because a prior `dnf install` already pulled `openssl-libs` forward — produces zero signal, and the script still prints its `"pinned to ..."` success line.

Why it matters: in the **test job**, "Install prereqs" runs `dnf -y install findutils git ...` against the floating UBI repos *before* the pin step, so `openssl-libs` can be floated up; this `distro-sync` is the only thing that pulls it back down to GA. The whole point of removing `dnf -y update` from the test job (promise 3) is that the test must run against the *old* GA `libcrypto` so a bad binary fails. If this sync silently fails, the test env keeps the newer OpenSSL, `asbackup --help` loads fine against it, and the regression backstop reports green while real RHEL 9.4 / 10.0 customers would fail. A backstop that can silently no-op is not a backstop. "All checks pass" doesn't exonerate this — a no-op'd sync passes by construction.

Fix: don't suppress it. Let `set -e` kill the job, and assert the post-condition so the pin is self-checking:
```
dnf -y distro-sync openssl openssl-libs
v="$(rpm -q --qf '%{VERSION}' openssl-libs)"
case "$v" in 3.2.*) echo "pin: openssl-libs $v" ;; *) echo "pin: expected 3.2.x, got $v" >&2; exit 1 ;; esac
```

.github/workflows/build-and-release.yml
```diff
@@ -627,10 +631,8 @@ jobs:
               ;;
             el*|amzn*)
               command -v dnf >/dev/null || yum install -y dnf
-              # install_deps_* runs `dnf -y update` before linking RPMs; without the same
-              # here, OpenSSL in a cold UBI/AL image can be older than libcrypto symbol
-              # versions the binaries were built against (e.g. OPENSSL_3.4.0).
-              dnf -y update
+              # No `dnf -y update`: the "Pin EL build repos" step makes this test run against
+              # the GA OpenSSL real customers have, so a refloated symbol floor would fail here.
```
dwelch-spike: From Tests agent: **Important:** This comment asserts the backstop catches a refloated symbol floor, but the only thing the test runs is `asbackup --help` / `--version` (`.github/bin/test/test_execute.bats`). That has two gaps for the exact bug this PR fixes:

1. The PR's failure log has **two** modes: a load-time version error (`version OPENSSL_3.4.0 not found`) *and* a lazy-bound `undefined symbol: EVP_MD_CTX_get_size_ex ... at S3 init`. `--help` triggers the dynamic linker's version check (so it would catch the first), but with lazy binding it never calls into the S3 path, so it would **not** catch the second symbol failure. The test passes on a binary that still dies the moment S3 is used.

2. Even for the version check, the test only proves "the binary loads against whatever `libcrypto` the pinned container has" — which is the same vault snapshot the build used. It's circular: it confirms build-env == test-env, not that the floor is ≤ `OPENSSL_3.2.0`. If the pin silently fails (see the `distro-sync` finding), build and test float together and every test stays green.

Fix: add a direct, container-independent floor assertion in the build job (binutils is present after compile), so a regression fails at build time, not at a customer:
```
! objdump -T asbackup | grep -Eo 'OPENSSL_3\.[0-9]+' | sort -u | grep -qvE 'OPENSSL_3\.[0-2]'
```
(fail if any referenced OpenSSL symbol version is > 3.2). The PR's "Testing" checkbox is also still unchecked — there's currently no evidence, direct or indirect, that the floor holds.

.github/bin/pin_build_repos.sh
```diff
@@ -0,0 +1,37 @@
...
+# Freeze EL9/EL10 to the Rocky vault GA snapshot for the base image's minor, so the
+# dynamically linked OpenSSL floor stays at the oldest the major shipped (3.2.2 ->
+# OPENSSL_3.2.0). The default UBI repos float to 3.5 (9.7/10.1), whose OPENSSL_3.4.0
+# symbols (e.g. EVP_MD_CTX_get_size_ex via the AWS SDK) the older RHEL 9.x/10.0
+# libcrypto can't satisfy. Run as a subprocess (not sourced); no-op on other distros.
+set -euo pipefail
+
+distro="${1:-${ENV_DISTRO:-}}"
+case "$distro" in
+  el9)  rel=9.6  ;;   # match the ubi9:9.6 / ubi10:10.0 base GA (no downgrade)
```
dwelch-spike: From Style agent: **Minor:** The `el9` branch's inline comment references `ubi10:10.0`, which has nothing to do with el9 — copy-paste residue that will mislead whoever edits this matrix next. The el10 branch below it (`el10) rel=10.0 ;;`) has no comment at all. Make each comment describe its own branch:
```
  el9)  rel=9.6  ;;   # match the ubi9:9.6 base GA (no downgrade)
  el10) rel=10.0 ;;   # match the ubi10:10.0 base GA (no downgrade)
```

.github/bin/pin_build_repos.sh
a-spiker: Good catch on both. Fixed in f68669b: switched to `gpgcheck=1` and pointed dnf at the Rocky release key over HTTPS (`dl.rockylinux.org/pub/rocky/RPM-GPG-KEY-Rocky-{9,10}`). CI confirms the key imports and packages verify against it. Dropped the misleading `(as aql/asadm do)` clause.

Also reworded the PR description: this pin is new to the suite (aql/asadm run `dnf -y update` and only pull individual vault RPMs by URL, they don't pin), and the reason asbackup needs it is the statically-linked AWS SDK (`aws-c-cal`) binding newer OpenSSL symbols those two never pull in.

.github/bin/pin_build_repos.sh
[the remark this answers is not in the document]
a-spiker: Fixed in f68669b. Each case arm now documents its own branch: el9 -> ubi9:9.6, el10 -> ubi10:10.0, both `openssl-libs` 3.2.2.

a-spiker: Fixed in f68669b. Each case arm now documents its own branch: el9 -> ubi9:9.6, el10 -> ubi10:10.0, both `openssl-libs` 3.2.2.

.github/bin/pin_build_repos.sh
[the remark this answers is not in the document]
a-spiker: Fixed. Dropped `2>/dev/null || true` so a real failure stops the job (f68669b). Making it fail-loud immediately surfaced a benign case the old `|| true` had been hiding: on el10 the `distro-sync` was forcing a same-version reinstall of `openssl-libs` (already at the 3.2.2 GA floor), which collides on `/usr/lib64/ossl-modules/fips.so` because the vault bundles that file while the ubi10 base splits it into `openssl-fips-provider-so`. 0c7f984 now only syncs when `openssl-libs` has actually floated past the 3.2 floor, then asserts the result, so a genuine sync failure still fails the job.

a-spiker: Fixed. Dropped `2>/dev/null || true` so a real failure stops the job (f68669b). Making it fail-loud immediately surfaced a benign case the old `|| true` had been hiding: on el10 the `distro-sync` was forcing a same-version reinstall of `openssl-libs` (already at the 3.2.2 GA floor), which collides on `/usr/lib64/ossl-modules/fips.so` because the vault bundles that file while the ubi10 base splits it into `openssl-fips-provider-so`. 0c7f984 now only syncs when `openssl-libs` has actually floated past the 3.2 floor, then asserts the result, so a genuine sync failure still fails the job.

.github/workflows/build-and-release.yml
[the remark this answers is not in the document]
a-spiker: Reworded in f68669b to make the order explicit: this install runs before checkout against the default UBI repos (so it can float `openssl-libs` forward), and the later "Pin EL build repos" step distro-syncs it back down to the GA OpenSSL customers have.

a-spiker: Agreed, the install test alone is circular and misses the lazily-bound S3 symbol path. Added `.github/bin/assert_openssl_floor.sh` (f68669b): an `objdump -T` check on the freshly built asbackup/asrestore that fails the build if any referenced OpenSSL symbol version is above 3.2, independent of whatever libcrypto the container has. Wired into the build job (el9/el10) and build-check (el10 per-commit), so a refloat fails at build time instead of at a customer.

a-spiker:
Thanks, the cross-repo dig was the most useful part. Fixes for all five gates are pushed:

- **Fail-open sync**: fail-loud now, and only syncs when OpenSSL actually floated, with a post-condition assert.
- **gpgcheck**: `gpgcheck=1` with the Rocky key over HTTPS instead of disabled.
- **Symbol-floor evidence**: new `assert_openssl_floor.sh` `objdump` backstop in both build and build-check, so the floor is proven directly rather than via build-env == test-env.
- **Comments**: el9/el10 split apart, install/pin ordering clarified, the false `(as aql/asadm do)` line removed.
- **Cross-repo / promise 5**: you're right, this pin is new to the suite, not an aql/asadm convention. Reworded the PR description to say so and to explain the AWS SDK (`aws-c-cal`) reason asbackup specifically needs it. Also called out the el9-vs-el10 per-commit gap (build-check only runs el10; el9 is covered by the full release workflow).

dwelch-spike (approved):
LGTM

PR #182, opened 2026-09-10
feat(ci): TOOLS-4364 build docker images via shared-workflows

a-spiker:
Ports the docker CI rework from aerospike-admin (aerospike/aerospike-admin#552) so all six tools repos share one pipeline shape.

## What changes

`docker-build-arch` + `docker-manifest` are replaced by three jobs:

| job | when | what it does |
|---|---|---|
| `docker-verify` | every dispatch | Builds each platform natively from the signed `.deb` and smoke-tests it. Publishes nothing. |
| `docker-build`  | every dispatch, pushes only on release | Delegates build / push / manifest / build-info / attest to `reusable_docker-build-deploy@v4.1.0`. |
| `docker-test`   | release only | Pulls the published index on each arch and confirms it is the build that passed `docker-verify`. |

`docker-verify` gates `docker-build`, so a broken image can no longer reach dev-local: the shared workflow pushes from inside its own build leg and offers no test hook, which is why the gate has to be a separate job.

Also in this PR:

- **Per-leg JFrog build-info** in `build-artifacts` (environment + git revision per distro/arch). Best-effort: a leg that fails here still ships, and warns.
- **`attest-artifacts`**: SLSA provenance for the signed packages. Runs after signing, because signing rewrites the deb/rpm in place and an attestation of the unsigned digest would never verify.
- **`.dockerignore`**: the build context carries only `deploy-artifacts`, instead of the whole working tree including `.git`.
- **Dockerfile**: fails loudly when the named `.deb` is missing from the build context, rather than silently falling back to an unverified download.
- **`setup-jfrog-cli` bumped to v5.1.0** (Node 24 runtime).

## Breaking: per-arch image tags go away

`reusable_docker-build-deploy` pushes each platform *by digest* and tags only the manifest list, so `<release>-amd64` and `<release>-arm64` are no longer published. Docker Hub keeps the multi-arch index and its aliases; only the per-arch tags are dropped.

Updating `copy-tools-images-to-dockerhub.yml` to stop defaulting to those tags is a **companion PR** on `TOOLS-4364-dockerhub-tags`, kept separate so this PR touches only the build pipeline. The two should land together: merge this one alone and the copy workflow still defaults to tags that no longer exist.

## Notes

- Pinned to `v4.1.0`, the current shared-workflows release. `gh-context-artifacts-json`, which puts the signed `.deb` inside the build context, only exists from v4.1.0; without it the image could only install a package downloaded from a URL rather than the artifact the run just signed.
- `versions-override` is passed to keep `latest` and the bare major from moving on an rc build. shared-workflows after v4.1.0 (INFRA-727) drops those from the default tag set, at which point the override becomes belt-and-braces; it is kept so the bump is a no-op.
- `jf-build-names` versions the docker build-info by `github.run_number` and the package build-info by `VERSION`, on purpose: the shared workflow records images under the JFrog CLI default build number. This matches the reference `example_docker-test.yaml`.

## Verification

- `actionlint` clean; finding counts compared against each repo's baseline, so nothing new was introduced.
- Job graph and every shared step diffed against the other five repos after normalizing tool/package/OIDC names: `docker-build` and `attest-artifacts` are byte-identical across all six. `docker-verify` and `docker-test` differ only in which binaries get asserted and whether `astools.conf` is checked, which tracks real product differences.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+# docker/Dockerfile builds with the repo root as its context so CI can drop the
+# signed .deb at deploy-artifacts/ (see the docker-build job). The image reads
+# that one path and nothing else, so everything else is excluded -- otherwise
+# every build leg uploads the whole working tree, .git included.
+# Re-include here anything a future COPY or bind-mount read needs: an excluded
+# path is simply absent from the context, which the Dockerfile reports as
+# "not in the build context" for the .deb and would silently miss elsewhere.
+*
+!deploy-artifacts
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -155,6 +155,8 @@ jobs:
       VERSION: ${{ steps.compute.outputs.version }}
       BUILD_VERSION: ${{ steps.compute.outputs.build_version }}
       RELEASE: ${{ steps.release.outputs.RELEASE }}
+      DEB_AMD64: ${{ steps.deb-name.outputs.amd64 }}
+      DEB_ARM64: ${{ steps.deb-name.outputs.arm64 }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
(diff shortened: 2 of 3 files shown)


pvinh-spike (approved):
LGTM

